PR #219, opened 2015-01-28
Fix/appscan files as objects

geoffwhittington:


sdetools/analysis_integration/tests/base_integration_test.py
```diff
@@ -80,6 +82,32 @@ def test_import_with_multiple_files(self):
 
         self.test_import_findings_assert_failed_tasks()
 
+    def _test_import_with_zip(self):
```
emintham: Does the test discovery run this test run with the underscore in front of it?

geoffwhittington: no, only "^test" methods are run

funkaoshi:
+1

PR #222, opened 2015-02-09
Allow users to bulk search and replace text in project task notes

geoffwhittington:
This command was motivated first by a need to rewrite ALM ticket references if an ALM server name is changed. It is generic enough to be used for other purposes


geoffwhittington:
This is a command-line only script which is not targeted for use by anyone

PR #223, opened 2015-02-10
Sync tasks based on verification status

geoffwhittington:


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -554,9 +563,16 @@ def in_scope(self, task):
 
         in_scope = (task['phase'] in self.config['alm_phases'] and
                     task['priority'] >= self.config['sde_min_priority'])
+        if not in_scope:
+            return False
 
         if self.config['sde_tags_filter']:
```
funkaoshi: this is 100% style, and i'm not sure its any nicer, but another way to do this is something like:

```
in_scope = (task['phase'] in self.config['alm_phases'] and
            task['priority'] >= self.config['sde_min_priority'])
 
if in_scope and self.config['sde_tags_filter']:
    in_scope &= set(self.config['sde_tags_filter']).issubset(task['tags'])
  
if in_scope and self.config['sde_verification_filter']:
    in_scope &= task['verification_status'] in self.config['sde_verification_filter']

return in_scope
```

Or something like that anyway. Manually short circuiting the function looks awkward to my eyes, but I don't know if this approach is any nicer.

funkaoshi:
This looks good. 👍

sdetools/alm_integration/alm_plugin_base.py
geoffwhittington: At one point I had something similar but opted to save a few cycles and wrote what is there now. But I do like the aesthetics of your solution and it feels nicer.

PR #225, opened 2015-02-13
Sanity check the weakness-task mappings

geoffwhittington:
Add capability to sanity check analysis weakness-task mappings against the capabilities listed in the task library


geoffwhittington:
This is blocking on a necessary API improvement

geoffwhittington:
I've updated the weaknesses in this PR at the same time

sdetools/sdelib/sdeapi.py
```diff
@@ -169,5 +169,23 @@ def get_phases(self, options={}, **filters):
         """
         return self.call_api('phases', args=filters, call_headers=_encode_options(options))
 
+    def get_library_tasks(self, options={}, **filters):
```
funkaoshi: Shouldn't use {} as the default argument in a function in python, because you can end up with some weird behaviour.

http://effbot.org/zone/default-values.htm

sdetools/analysis_integration/base_integrator.py
```diff
@@ -554,6 +666,9 @@ def mapping_contains_task(self, needle_task_id):
         return False
 
     def import_findings(self):
+        if self.config['check_coverage']:
```
funkaoshi: Is there a better place to abort things if we are just checking for coverage?

sdetools/analysis_integration/base_integrator.py
```diff
@@ -497,6 +509,106 @@ def load_mapping_from_xml(self):
...
+                continue
+
+            task_id = task_search.group(1)
+
+            title_search = re.search('^T\d+:\s*(.+)$', library_task['title'])
+            if title_search:
+                task_title = title_search.group(1)
+            else:
+                task_title = None
+
```
funkaoshi: Extra blank line.

sdetools/analysis_integration/base_integrator.py
geoffwhittington: The check can either happen here or by the caller:

https://github.com/sdelements/sdetools/blob/master/sdetools/modules/import_whitehat/__init__.py

```
 def handle(self):
        self.wh_integrator.initialize()
        self.wh_integrator.load_mapping_from_xml()
        self.wh_integrator.parse()
        self.emit.info('Finding file parsed successfully. Starting the import')
        self.wh_integrator.import_findings()
```

geoffwhittington: Might be good to have the caller do it

sdetools/analysis_integration/base_integrator.py
```diff
@@ -497,6 +509,106 @@ def load_mapping_from_xml(self):
...
+        for library_task in library_tasks:
+            # We don't support custom tasks at this moment
+            task_search = re.search('^T(\d+)$', library_task['id'])
+            if not task_search:
+                continue
+
+            task_id = task_search.group(1)
+
+            title_search = re.search('^T\d+:\s*(.+)$', library_task['title'])
+            if title_search:
```
funkaoshi: You can write this as: `task_title = title_search.group(1) if title_search else None`. People also sometimes write this as `task_title = title_search and title_search.group(1)`.

sdetools/analysis_integration/base_integrator.py
```diff
@@ -405,6 +408,10 @@ def set_tool_name(self, tool_name):
         self.TOOL_NAME = tool_name
 
     def parse(self):
+        if self.config['check_coverage']:
```
funkaoshi: What calls parse? Maybe we can do this call before that?

geoffwhittington:
We're packing a lot of changes into this PR. Instead I'm moving the custom mapping support to #287 The rest of the changes will get moved to other PR's on a need basis

PR #226, opened 2015-02-17
Allow users to filter based on no verification status

geoffwhittington:


geoffwhittington:
This fixes the current code to include no verification status

PR #235, opened 2015-03-23
Fix suds import error

geoffwhittington:


sdetools/extlib/__init__.py
```diff
@@ -5,4 +5,9 @@
     'argparse_compat',
     'markdown',
     'SOAPpy',
+    'suds',
     ]
+
+import os
+import sys
+sys.path.append(os.path.dirname(os.path.realpath(__file__)))
```
funkaoshi: It's probably not best to futz with the python path in your code like this. See what other people think. I think safer change is update imports in the code to import from `sdetools.extlib.suds`.

geoffwhittington: We are already futzing with the python path, actually. The problem is suds assumes its already in the python path.

funkaoshi: Might want to remove references to imports from extlib then in the rest of the code base, they won't be needed. (And might want to comment saying we are adding extlib to the python path.)

PR #237, opened 2015-03-27
Update for Android

geoffwhittington:


geoffwhittington:
Verified that imports work with this new XML and mappings were reviewed internally

PR #238, opened 2015-03-27
Rally tag fix: truncate tags to 32 characters

geoffwhittington:


geoffwhittington:
@ehsanf Fix is in

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -332,6 +335,10 @@ def rally_get_artifact(self, name, query, card_type, artifact_type, api):
             raise AlmException('Unable to get artifact from Rally. Reason: %s' % (str(err)))
         return task_data[artifact_type]
 
+    def _transform_label_to_rally_tag(self, label):
+        """ Rally supports tags having at most 32 characters """
+        return label[0:32]
```
funkaoshi: `return label[:32]` is probably more pythonic? People usually drop the starting zero if you are slicing from the start of the list.

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -213,6 +213,9 @@ def alm_connect_project(self):
 
     def _setup_rally_label(self, tag_name):
 
+        # Ensure tags are transformed
```
funkaoshi: ` ... because Rally only supports 32 character tags`

Should say why we need to do this.

funkaoshi:
+1

PR #241, opened 2015-04-19
Sanity test done statuses on startup and allow mixed case done statuses

geoffwhittington:


sdetools/modules/sync_jira/jira_soap.py
```diff
@@ -306,6 +306,14 @@ def get_available_transitions(self, alm_id):
             ret_trans[transition['name']] = transition['id']
         return ret_trans
 
+    def get_statuses(self):
+        statuses = []
```
sibartlett: Use a list comprehension here.

sdetools/modules/sync_jira/jira_plugin.py
```diff
@@ -106,6 +106,16 @@ def alm_validate_configurations(self):
         if missing_priorities:
             raise AlmException('Incorrect priority mapping values specified: %s' % ', '.join(missing_priorities))
 
+        available_statuses = self.alm_plugin.get_statuses()
+        invalid_statuses = []
+        for status in self.config['jira_done_statuses']:
+            if status.lower() not in [available_status.lower() for available_status in available_statuses]:
```
funkaoshi: You are recomputing the lower case list of available statuses each time you go through this loop.

```
lowercased_available_statuses = [available_status.lower() for available_status in available_statuses]
invalid_statuses = [status for status in self.config['jira_done_statuses'] if status.lower in lowercased_available_statuses]
```

PR #242, opened 2015-04-21
Fix tests for Jira 5.x support

geoffwhittington:
Jira 5.x API does not support `/rest/project/KEY/statuses`. We will have to treat it like SOAP and validate against all available statuses


funkaoshi:
+1

PR #243, opened 2015-04-23
Support business units

geoffwhittington:


sdetools/sdelib/testlib/sde_response_generator.py
```diff
@@ -94,6 +100,21 @@ def get_taskstatuses(self, target, flag, data, method):
         else:
             self.raise_error('401')
 
+    def call_businessunits(self, target, flag, data, method):
+        if not flag:
```
funkaoshi: I usually re-write stuff like this to decrease the amount of indentation needed.

```
if flag:
    self.raise_error('401')

if method != 'GET':
    self.raise_error('400')

params = self.get_url_parameters(target)
...
...
```

PR #246, opened 2015-05-15
Feature/import csv

jnlantos:


sdetools/modules/import_csv.py
```diff
@@ -0,0 +1,70 @@
...
+
+        with open(self.config['csv_file'], 'r') as csvfile:
+            reader = csv.DictReader(csvfile)
+            
+            if reader.fieldnames:
+                # Remove extra white space in headings
+                for i in range(len(reader.fieldnames)):
+                    reader.fieldnames[i] = reader.fieldnames[i].strip()
+
+                if len(reader.fieldnames) <= 1:
```
geoffwhittington: we should check for "task id" field explicitly

sdetools/modules/import_csv.py
```diff
@@ -0,0 +1,70 @@
...
+                    raise Error('Please make sure CSV file has task id field and at least one more field.')
+            else:
+                raise Error('Empty CSV file')
```
geoffwhittington: Process the `else` before the `if` so you can avoid indenting code

```
if not reader.fieldnames:
    raise Error('Import CSV empty")

```

sdetools/modules/import_csv.py
```diff
@@ -0,0 +1,70 @@
...
+    def configure(self):
+        self.sde_plugin = PlugInExperience(self.config)
+        self.config.opts.add('csv_file', "CSV file")
+
+    def sde_connect(self):		
+        if not self.sde_plugin:
+            raise Error('Requires initialization')
+        try:
+            self.sde_plugin.connect()
+        except APIError, err:
```
emintham: This should probably be `except APIError as err` instead since we only support 2.6 and above. https://www.python.org/dev/peps/pep-3110/#compatibility

sdetools/modules/import_csv.py
```diff
@@ -0,0 +1,70 @@
...
+                        self.sde_plugin.add_task_text_note(task_id, text_note)
+                    if 'verification status' in reader.fieldnames:
+                        verification_status = row['verification status']
+                        if 'verification note' in reader.fieldnames:
+                            finding_ref = row['verification note']
+                        else:
+                            finding_ref = ''
+                        self.sde_plugin.add_manual_analysis_note(task_id, verification_status, finding_ref)
+
+            except KeyError, key_err:
```
emintham: As above.

sdetools/modules/import_csv.py
```diff
@@ -0,0 +1,70 @@
...
+                raise Error('Unable to parse CSV file. Please review column headings. Reason: %s' % (str(key_err)))
+
+            except csv.Error, csv_err:
```
emintham: As above.

sdetools/sdelib/sdeapi.py
```diff
@@ -142,6 +142,20 @@ def get_task_notes(self, task, note_type, options={}, **filters):
         args.update(filters)
         return self.call_api(end_point, args=args, call_headers=_encode_options(options))
 
+    def add_manual_analysis_note(self, task, verification_status, finding_ref):
+        if verification_status == "pass" or verification_status == "fail":
```
emintham: You can do `if verification_status in ['pass', 'fail']:` instead. Or, better yet, make that a variable in the class e.g. `HIGH_CONFIDENCE_STATUSES = ['pass', 'fail']` and do `if verification_status in self.HIGH_CONFIDENCE_STATUSES: confidence = 'high'`. If the strings that indicate the verification statuses change, it will be easier to find them and change it in just one place.

In fact, you can also make it a separate method to make this method cleaner.
```python
def get_confidence(self, verification_status):
    if verification_status in self.HIGH_CONFIDENCE_STATUSES:
        return 'high'
    elif verification_status in self.LOW_CONFIDENCE_STATUSES:
        return 'low'
```

sdetools/modules/import_csv.py
geoffwhittington: Actually, we try to support 2.4+ in sdetools.

sdetools/sdelib/sdeapi.py
geoffwhittington: Good feedback. We can simply further: just return "low" with an else and drop the elif, and the low confidence variable.

sdetools/modules/import_csv.py
emintham: Ah, my bad. I didn't realise that-- I just assumed the README was outdated and took the travis versions to be the official ones.

sdetools/sdelib/sdeapi.py
emintham: Good point.

sdetools/modules/import_csv.py
geoffwhittington: No worries, looks like we will have to test 2.4 ourselves using tox, most likely

sibartlett: @emintham Travis doesn't support Python 2.4 :cry:

PR #254, opened 2015-05-26
Feature/command driver

geoffwhittington:
- [x] Squash commits


sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,35 @@
...
+            raise Error('Unable to connect to SD Elements. Please review URL, id,'
+                    ' and password in configuration file. Reason: %s' % (str(err)))
+
+    def handle(self):
+        result = self.sde_plugin.get_integration_jobs()
+        jobs = result['result']
+
+        for job in jobs:
+            command = job['command']
+            args = job['params']
```
geoffwhittington: Call this `params` to be consistent and avoid confusion with `args` in the `sdetools.call` parameter

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,35 @@
...
+        if not self.sde_plugin:
+            raise Error('Requires initialization')
+        try:
+            self.sde_plugin.connect()
+        except APIError, err:
+            raise Error('Unable to connect to SD Elements. Please review URL, id,'
+                    ' and password in configuration file. Reason: %s' % (str(err)))
+
+    def handle(self):
+        result = self.sde_plugin.get_integration_jobs()
```
geoffwhittington: rename `result` as `response`

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,35 @@
...
+from sdetools.sdelib.interactive_plugin import PlugInExperience
+import sdetools
+
+
+class Command(BaseCommand):
+    help = 'Downloads project integration job details from SDE and executes the jobs'
+    sde_plugin = None
+
+    def configure(self):
+        self.sde_plugin = PlugInExperience(self.config)
```
geoffwhittington: The `PlugInExperience` is probably too much here (we don't need business units, application or projects to be defined) - may as well just use `sdeapi.APIBase`

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+
+class Command(BaseCommand):
+    help = 'Downloads project integration connection details from SDE and executes them as jobs'
+
+    def configure(self):
+        self.api = APIBase(self.config)
+        self.job_return_msgs = {}
+        self.config.opts.add('connection_name', "Connection name")
+        self.config.opts.add('frequency', "Frequency (manual|hourly|daily|weekly|monthly)")
+        self.config.opts.add('connection_type', "Alm connector type")
```
geoffwhittington: `Alm connector type` => `Connector type (ALM or Analysis Tool`

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+    def sde_connect(self):
+        if not self.api:
+            raise Error('Requires initialization')
+        try:
+            self.api.connect()
+        except APIError, err:
+            raise Error('Unable to connect to SD Elements. Please review URL, id,'
+                    ' and password in configuration file. Reason: %s' % (str(err)))
+
+    def filter(self, connections):
```
geoffwhittington: Can you change the method to something like `_filter_connections` to denote that it is internal (_) and to better describe what it is doing?

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+        filtered_connections = []
+
+        for connection in connections:
+            if self.config['connection_name'] and connection['name'] != self.config['connection_name']:
+                continue
+            if self.config['frequency'] and connection['frequency'] != self.config['frequency'] :
+                continue
+            if self.config['connection_type'] and connection['type'] != self.config['connection_type']:
+                continue
+            if self.config['connection_ids'] and str(connection['id']) not in self.config['connection_ids'].split(","):
```
geoffwhittington: We should validate these inputs in a separate method so we can be assured that these id's are numbers and not something else?

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def filter(self, connections):
+        filtered_connections = []
+
+        for connection in connections:
+            if self.config['connection_name'] and connection['name'] != self.config['connection_name']:
+                continue
+            if self.config['frequency'] and connection['frequency'] != self.config['frequency'] :
+                continue
+            if self.config['connection_type'] and connection['type'] != self.config['connection_type']:
```
geoffwhittington: We should validate this value in a separate method - similar to what we do here: https://github.com/sdelements/sdetools/blob/master/sdetools/alm_integration/alm_plugin_base.py#L399

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+                continue
+            if self.config['connection_ids'] and str(connection['id']) not in self.config['connection_ids'].split(","):
+                continue
+            if self.config['params']:
```
geoffwhittington: we should validate that the params has valid entries or we risk failing on invalid input like the above comments

sdetools/modules/command_driver.py
geoffwhittington: Can you rewrite to match https://github.com/sdelements/sdetools/blob/master/sdetools/alm_integration/alm_plugin_base.py#L555

and

https://github.com/sdelements/sdetools/blob/master/sdetools/alm_integration/alm_plugin_base.py#L645

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+                        # break - slow speed up if the user put a bunch of param filters...not really worth it maybe
+                if not match_all_params:
+                    continue
+
+            filtered_connections.append(connection)
+
+        return filtered_connections
+
+    def on_job_complete(self, log_info, id):
+        self.job_return_msgs.update({id:log_info.msg})
```
geoffwhittington: is there anything else in `log_info` we might want to use later? Wondering we can add the whole object instead of just the `.msg`

sdetools/sdelib/sdeapi.py
```diff
@@ -45,10 +45,19 @@ def connect(self):
         else:
             #TODO: Find a better alternative ->
             #In 'basic' mode, we make an extra call just to verify that credentials are correct
-            result = self.get_applications()
+            result = self.get_users()
         self.connected = True
         return not not result
 
+    def get_users(self, myself=False):
```
geoffwhittington: What do you think of `myself_only`?

sdetools/sdelib/sdeapi.py
```diff
@@ -199,5 +208,19 @@ def get_phases(self, options={}, **filters):
         """
         return self.call_api('phases', args=filters, call_headers=_encode_options(options))
 
+    def get_integration_connections(self):
+        """
+        Get project integration job details
+        """
+        return self.call_api('projectintegrationconnections')
+
+    def update_job_status(self, status, msg, id):
```
geoffwhittington: We need to send the `type` as a parameter since we will have Analysis and ALM type job statuses.

sdetools/sdelib/sdeapi.py
```diff
@@ -199,5 +208,19 @@ def get_phases(self, options={}, **filters):
...
+       job_status = {
+            'type': "ALM",
```
geoffwhittington: Replace `"ALM"` with `type`

sdetools/sdelib/sdeapi.py
geoffwhittington: Rename `id` to `connection_id` and move it to after `self` since it is the most important parameter

sdetools/sdelib/testlib/sde_response_generator.py
```diff
@@ -86,6 +88,33 @@ def get_phases(self, target, flag, data, method):
...
+        if not flag:
+            if method == 'GET':
+                user = {
+                    'first_name': 'First Name',
+                    'last_name': 'Last Name',
+                    'email': 'email@example.com'
+                }
+
+                if myself_only:
+                    return RESPONSE_HEADERS, {
```
geoffwhittington: Should be `return RESPONSE_HEADERS, user`

sdetools/sdelib/sdeapi.py
```diff
@@ -45,10 +45,19 @@ def connect(self):
         else:
             #TODO: Find a better alternative ->
             #In 'basic' mode, we make an extra call just to verify that credentials are correct
-            result = self.get_applications()
+            result = self.get_users()
```
geoffwhittington: Nice. This will save some time and traffic

geoffwhittington: Send in the `myself` flag so that we only retrieve the user's information, instead of everyone in the organization

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+            exit_stat = sdetools.call(command, options=params, call_back=self.on_job_complete, call_back_args={"id":connection_id})
+            if not exit_stat:
+                failed_jobs.append(connection_id)
+            else:
+                successful_job_count += 1
+            self.api.update_job_status(exit_stat, self.job_return_msgs[connection_id], connection_id)
+
+        logger.info('%s jobs completed successfully, %s failed' % (successful_job_count, len(failed_jobs)))
+
+        if len(failed_jobs) > 0:
```
geoffwhittington: you can simplify with `if failed_jobs:`

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+            logger.info('Failed jobs by id: %s' % failed_jobs)
```
geoffwhittington: Format the job id's: `', '.join(failed_jobs)` in the information message

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+        self.config.opts.add('params', "Filter connections based on params such as sde_businessunit or sde_application (key1:value1,key2:value2...)")
+        self.config.opts.add('connection_ids', "Connection ids (1,2,7...)")
+
+    def sde_connect(self):
+        if not self.api:
+            raise Error('Requires initialization')
+        try:
+            self.api.connect()
+        except APIError, err:
+            raise Error('Unable to connect to SD Elements. Please review URL, id,'
```
geoffwhittington: Might be better to refer to say `Please review URL, and credentials in configuration...` since its more likely that people will use their api connection string than a user/password

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+from sdetools.sdelib import log_mgr
+logger = log_mgr.mods.add_mod(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Downloads project integration connection details from SDE and executes them as jobs'
+
+    def configure(self):
+        self.api = APIBase(self.config)
+        self.job_return_msgs = {}
```
geoffwhittington: `job_return_args` is better defined inside of the class or inside an `__init__` method.

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+        self.config.opts.add('connection_name', "Connection name")
+        self.config.opts.add('frequency', "Frequency (manual|hourly|daily|weekly|monthly)")
+        self.config.opts.add('connection_type', "Alm connector type")
+        self.config.opts.add('params', "Filter connections based on params such as sde_businessunit or sde_application (key1:value1,key2:value2...)")
```
geoffwhittington: You can simplify the help string as `Filter connections based on params (key1:value1,key2:value2...)`

geoffwhittington: I'm doubtful people will actually use this except for specific troubleshooting situations

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+        logger.info('Running %s jobs' % len(filtered_connections))
+
+        for connection in filtered_connections:
+            command = connection['command']
+            params = connection['params']
+            params['sde_api_token'] = self.config['sde_api_token']
+            params['sde_method'] = self.config['sde_method']
+            connection_id = connection['id']
+
+            exit_stat = sdetools.call(command, options=params, call_back=self.on_job_complete, call_back_args={"id":connection_id})
```
geoffwhittington: Can we add a `logger.info` statement saying which job is running (connection id, name, type)

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,95 @@
...
+        failed_jobs = []
+
+        filtered_connections = self.filter(connections)
+        logger.info('Filtered out %s connections' % (len(connections) - len(filtered_connections)))
+        logger.info('Running %s jobs' % len(filtered_connections))
+
+        for connection in filtered_connections:
+            command = connection['command']
+            params = connection['params']
+            params['sde_api_token'] = self.config['sde_api_token']
```
geoffwhittington: Add a comment saying why we're doing this - basically we are setting the credentials and method with which to connect to SDE, we don't get this from the server.

sdetools/modules/command_driver.py
geoffwhittington: I think you mean `exit_status` ?

geoffwhittington:
There's a couple of things to address:
- Exceptions are output without any project integration context
- See if we can log JUST the command_driver without also logging everything it ultimately runs

jnlantos:
Hmm, I'm not sure I understand exactly what you mean by 'project integration context'?

geoffwhittington:
I think you can hit the 403 by trying to sync to a project for which a user doesn't have access

sdetools/modules/command_driver.py
```diff
@@ -0,0 +1,129 @@
...
+            params = connection['params']
+            # Set the credentials and method with which to connect to SDE (we don't get this from the server)
+            params['sde_api_token'] = self.config['sde_api_token']
+            params['sde_method'] = self.config['sde_method']
+            connection_id = connection['id']
+            connector_type = connection['type']
+            connection_name = connection['alias']
+
+            logger.info('Running job with Connection Id: %s, Name: %s, Type: %s' % (connection_id, connection_name, connector_type))
+            exit_status = sdetools.call(command, options=params, call_back=self.on_job_complete, call_back_args={"connection_id":connection_id})
```
geoffwhittington: We should wrap this in a try/except - check for `AlmException` and `APIError`. I think that covers it....

geoffwhittington:
Due to the large number of commits, we'll want to squash all of them into a single commit. I'll take care of this

PR #256, opened 2015-05-26
Feature/verify analysis mappings

wisznie:
Cleaned up the request from earlier


sdetools/analysis_integration/base_integrator.py
```diff
@@ -471,32 +478,146 @@ def process_report_file_config(self):
...
+                if task not in csv_mapping[weakness]:
+                    errors.append("Mapping: Task '%s' => weakness '%s' not found in CSV" % (task, weakness))
+
+        return errors
+
+    def check_coverage(self):
+        """
+        Load the weakness->task mapping and sanity check against the task verification capabilities
+        """
+        self.load_mapping_from_xml()
```
geoffwhittington: We don't want to `check_coverage()` every time an import runs - only when the user asks to do it via the option, `self.config['check_coverage']`. We can check this option either before this method is called or inside this method. If you add a third option to this setting and mark is default: `bypass` or `nocheck` then you could use this as a way to jump out of the `check_coverage` method.

geoffwhittington: If `load_mapping_from_xml()` is called before `check_coverage()` is called then we don't need to call `self.load_mapping_from_xml()` inside `check_coverage()` - or do we?

PR #257, opened 2015-05-27
Feature/verify analysis mappings

wisznie:


geoffwhittington:
How is this different from #256 ?

wisznie:
Ah, that;s my mistake, they are the same, just different branches. I can close this pull request or the other.

sdetools/analysis_integration/base_integrator.py
```diff
@@ -478,20 +478,13 @@ def process_report_file_config(self):
...
-        try:
-            base = minidom.parse(self.config['mapping_file'])
-        except KeyError:
-            raise IntegrationError("Missing configuration option 'mapping_file'")
-        except Exception, e:
-            raise IntegrationError("An error occurred opening mapping file '%s': %s" % (self.config['mapping_file'], e))
-
+    def _load_mapping_from_xml(self, xml_file):
         weakness_mapping = collections.defaultdict(list)
         self.weakness_title = {}
```
geoffwhittington: self.weakness_title, self.confidence, self.weakness_type, self.task_titles must persist, we can't empty their values when the custom mapping is loaded

geoffwhittington: Maybe move these initialization statements somewhere safer?

geoffwhittington:
Can you close either this pull request or the other? I can't keep track of which one I should review :)

sdetools/analysis_integration/base_integrator.py
wisznie: Maybe I misunderstand how it's working, but the private function loads all the values from the initial setting file and returns a dictionary with all the values from self.weakness_title, self.confidence, self.weakness_type, self.task_titles. Then the private function is called with the custom mapping file and updates it's results(possibly overwriting defaults). Thus all the values in self.weakness_title, self.confidence, self.weakness_type, self.task_titles aren't lost. They're saved in self.mapping.

geoffwhittington: The trouble with the method as-is, which we should probably revisit later, is that it returns the mapping but it also updates internal state that is used later. Its not the best thing, I agree. Search for `self.weakness_title` - you'll see reference to it later in the code. All these variables are used elsewhere in the class and are initialized here, in addition to the mapping. If you call the method twice we lose the previous values.

PR #265, opened 2015-06-12
Get alm task based on custom fields mapping OR custom lookup fields

jnlantos:
This fix addresses bug SDE-1172.


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -708,7 +712,9 @@ def synchronize(self):
 
             if self.config['start_fresh']:
                 for task in tasks:
-                    alm_task = self.alm_get_task(task)
+                    # Probably want to delete ALL tasks in the alm and want to lookup without any restrictions on the custom fields
```
jnlantos: This might be a problem.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -326,7 +326,11 @@ def alm_connect_project(self):
         pass
 
     @abstractmethod
-    def alm_get_task(self, task):
+    def exists_in_alm(self, task):
```
geoffwhittington: I don't believe we need a new method for checking for the existence of a task in an alm. We should be able to simplify the code: `exists_in_alm` should be true if and only if `alm_get_task` returns `None`.

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -385,11 +385,29 @@ def alm_get_task(self, task):
...
 
-        for field, value in self.config['alm_custom_lookup_fields'].items():
+        task_data = self.rally_get_artifact(task_id, '%s' % artifact_query, card_type_details['type'],
+                                            card_type_details['type'], card_type_details['api'])
+        if not task_data:
+            return False
+
+        return True
+
+    def alm_get_task(self, task, alm_custom_fields_type):
```
geoffwhittington: This is the only method I believe we should be modifying to address the Task sync issue when using custom lookup fields. If we are unable to retrieve a Task when the sync is configured with custom lookup fields - then its due to a faulty `artifact_query`.

Whenever custom lookup fields are used then a Task is found using both `self.config['alm_custom_lookup_fields']` and the `self.config['alm_fixed_title']`. The values in `self['alm_custom_fields']` are only used to assign values to an artifact (Story, Task, Defect) - never used to find artifacts.

geoffwhittington:
We should try to keep merge requests focused on the specific changes being proposed

jnlantos:
It has been determined that SDE-1172 is not actually a bug but expected behaviour due to invalid input.

PR #276, opened 2015-07-09
Fix the cert validation logic

geoffwhittington:


sdetools/extlib/http_req.py
```diff
@@ -233,10 +233,9 @@ def get_http_handler(mode, debuglevel=0, validate_certs=True):
             return urllib2.HTTPSHandler(debuglevel=debuglevel)
         else:
             if validate_certs:
-                ca_certs_store = CA_CERTS_FILE
+                return VerifiedHTTPSHandler(debuglevel=debuglevel, ca_certs=CA_CERTS_FILE)
             else:
-                ca_certs_store = None
-            return VerifiedHTTPSHandler(debuglevel=debuglevel, ca_certs=ca_certs_store)
+                return urllib2.HTTPSHandler(debuglevel=debuglevel)
```
funkaoshi: Seems fine. You can probably re-write so you only return the HTTPSHandler handler in one place.

PR #280, opened 2015-07-15
Feature/jira field mapping

geoffwhittington:


geoffwhittington:
@funkaoshi Can you review this?

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -266,11 +266,26 @@ def init_custom_fields(self):
...
-        for required_macro in macros:
-            if required_macro not in mapping or not mapping[required_macro]:
-                raise AlmException("Missing value for configuration macro ${%s}" % required_macro)
-        return Template(value).substitute(mapping).strip()
+        """
+        Search for all macros in value and substitute them with their values
+        """
+
+        if isinstance(value, list):
+            new_list = []
```
funkaoshi: Should use list comprehensions when possible. It simplifies this whole block.

```
return [self._transform_value(v, mapping) for v in value]
```

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -643,14 +650,11 @@ def status_match(self, alm_status, sde_status):
...
-
-        return tasks
+        return [self.transform_task(task) for task in tasks if self.in_scope(task)]
 
-    @staticmethod
-    def add_alm_title(config, task):
-        task['alm_full_title'] = AlmConnector.get_alm_task_title(config, task, fixed=False)
-        task['alm_fixed_title'] = AlmConnector.get_alm_task_title(config, task, fixed=True)
+    def transform_task(self, task):
+        task['alm_full_title'] = self.get_alm_task_title(self.config, task, fixed=False)
```
funkaoshi: Not part of this change, but I think it'd have been better to just have two methods, `get_alm_full_title` and `get_alm_fixed_title`. (And maybe they both call some common method.) Seems weird to pass in fixed=True. Maybe I don't understand the details here, though. (At the very least, get_alm_task_title should probably have fixed=False be the default.)

sdetools/modules/sync_jira/jira_plugin.py
```diff
@@ -68,6 +78,21 @@ def initialize(self):
...
+        issue_template = {}
+        issue_template.update(JIRA_DEFAULT_ISSUE_TEMPLATE)
+
+        if self.config['alm_custom_fields']:
+            issue_template.update(self.config['alm_custom_fields'])
+
+        return issue_template
+
+    def get_new_issue(self, task):
+        new_issue = {}
```
funkaoshi: You can also write this like a list comprehension, though i'm not 100% sold on it being nicer:

```
return dict((field, self._transform_value(value, task)) for field, value in self.new_issue_template.iteritems())
```

sdetools/modules/sync_jira/jira_plugin.py
```diff
@@ -206,5 +230,37 @@ def alm_set_version(self, task, version):
...
 
-    def convert_markdown_to_alm(self, content, ref): 
+    def convert_markdown_to_alm(self, content, ref):
         return convert_markdown(content)
+
+    def init_custom_fields(self):
+        """ Custom fields are used differently in Jira so avoid default behaviour """
+        pass
+
+    def transform_task(self, task):
```
funkaoshi: Needs a doc string. "It takes a task and returns a dictionary representation in __macros so that ..."

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -227,12 +217,33 @@ def set_version(self, task, project_version):
...
+        try:
+            # Add a link back to SD Elements
+            remote_url = 'issue/%s/remotelink' % issue['key']
+            args = {"object": {"url": task['url'], "title": task['title']}}
+            self.call_api(remote_url, method=self.URLRequest.POST, args=args)
+            
+        except APIError:
+            # This is not a terrible thing, the user may not have permission
+            # and I don't think we should enforce that they do
+            pass
```
funkaoshi: May still want to log something? This might hide other APIErrors? Or can we pass on the more specific case they don't have permission?

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -136,19 +123,22 @@ def setup_fields(self, issue_type_id):
                         else:
                             raise AlmException('Field %s is not searchable' % key)
 
+        missing_fields = []
         for field_id, field_details in self.fields.iteritems():
-            if field_details['required'] and field_id not in assigned_fields:
+            if (field_details['required'] and not field_details['hasDefaultValue'] and
```
funkaoshi: I'd do one predicate per line, so:

if (field_details['required'] and
    not field_details['hasDefaultValue'] and
    field_id not in JIRARestAPI.BASE_FIELDS and
    field_details['name'] not in new_issue_template):
    ...
```

File: sdetools/modules/sync_jira/jira_rest.py [hunk 7]
```diff
@@ -241,42 +252,28 @@ def add_task(self, task, issue_type_id, project_version):
...
+        if self.config['alm_project_version']:
+            args['fields']['versions'] = [{'name': self.config['alm_project_version']}]
+
         unsupported_fields = []
 
-        for field_id, custom_field_value in self.custom_fields.iteritems():
+        for field_id, mapped_field in self.fields.iteritems():
 
-            mapped_field = self.fields[field_id]
+            if mapped_field['name'] in new_issue_fields:
```
funkaoshi: Style thing: I do the control flow stuff first:

```
if mapped_field['name'] not in new_issue_fields:
    continue

arg_value = new_issue_fields[mapped_field['name']]
schema = mapped_field['schema']
...
```

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -284,16 +281,25 @@ def add_task(self, task, issue_type_id, project_version):
...
                     field_values = []
 
-                # Accommodate more than one custom field value in this field
-                if isinstance(custom_field_value, list):
-                    for list_item in custom_field_value:
-                        field_values.append(list_item)
+                # Accommodate more than one value for this field
+                if isinstance(arg_value, list):
+                    for list_item in arg_value:
+                        # Labels must be sanitized
```
funkaoshi: Should probably make a method to avoid duplicating this block of code twice. (Can even define the function inside this function).

Also don't know if the more pythonic thing to do is try and treat the thing as a list, and fall back to it being a single item:

```
try:
    for list_item in arg_value:
         process(field_values, arg_value)
except TypeError:
    process(field_values, arg_value)
```

There is probably a (better?) pattern people use here.

sdetools/modules/sync_jira/jira_soap.py
```diff
@@ -244,11 +244,11 @@ def set_version(self, task, project_version):
     
         return True
 
-    def add_task(self, task, issue_type_id, project_version):
+    def add_task(self, new_issue_fields, task, issue_type_id, project_version):
         # Add task
         selected_priority = None
         for priority in self.priorities:
-            if priority['name'] == task['alm_priority']:
+            if priority['name'] == task['__macros']['task_priority_translated']:
```
funkaoshi: We discussed in person, but it seems like you should hide this stuff from the client code. One idea would be to make task a dictionary like object, add _macros as an additional attribute, and then override __getitem__ to get the behaviour you want.

geoffwhittington:
I think this is good to go now. I wish squash and merge

PR #285, opened 2015-07-30
Add project connector plugin functionality

hanifvirani:
Adds the plugin api for project connector plugins
along with a base plugin interface.


sdetools/connector_plugin/manager.py
```diff
@@ -0,0 +1,74 @@
+import imp
+import os
+
+
+class ProjectConnectorPluginManager(object):
```
hanifvirani: We can change the name if we decide to go with something like "integration" instead.

hanifvirani: Right now this is Project Connector specific. As we discussed today, we can later convert this into a more generic plugin framwork that can be used by things like custom reports, and possibly core sdetools  stuff too.

sdetools/connector_plugin/__init__.py
```diff
@@ -0,0 +1,2 @@
+from manager import *
```
funkaoshi: Might want to export the specific things we are exposing?

sdetools/connector_plugin/manager.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Manager class for Project Connector Plugins.
+
+    This class is used to register and fetch plugins.
+    """
+
+    _plugins = None
+    PLUGINS_FOLDER = '/docs/sde/connector-plugins'
```
funkaoshi: I'd make the manager something you instantiate, so you can pass this is. (And store the plugins for that manager within the object.) I think you'll need to do something like this anyway if we want to support different plugins, for example.

sdetools/connector_plugin/plugin.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @abstractmethod
+    def fetch_data(self, id):
+        """
+        Returns the data for the project associated with the given id.
+
+        This method is expected to return a dictionary containing key value
+        pairs of project attributes and their values.
+
+        TODO: Flesh the docstring more once final structure is agreed upon.
```
funkaoshi: Will need to be able to return error codes and messages, presumably, in addition to the actual data.

sdetools/connector_plugin/plugin.py
```diff
@@ -0,0 +1,76 @@
+from abc import ABCMeta, abstractmethod, abstractproperty
+from collections import namedtuple
+
+from .manager import ProjectConnectorPluginManager
```
emintham: Why are we doing a relative import here?

sdetools/connector_plugin/sample_plugin.txt
```diff
@@ -0,0 +1,41 @@
...
+    """
+    This is a a sample Project Connector plugin.
+    """
+
+    name = "My Sample Connector"
+    description = "Here is the description"
+
+    @property
+    def connection_params(self):
+        username = ConnectionParam(name="username",
```
emintham: I'd put these instantiations in the `__init__` and just return those in the property. Seems weird for the property to be doing any instantiations.

geoffwhittington:
Just a reminder that we try to support Python 2.4+ in sdetools

sdetools/connector_plugin/sample_plugin.txt
```diff
@@ -0,0 +1,41 @@
...
+                                   description="This is a username",
+                                   label="User Name")
+
+        password = ConnectionParam(name="password",
```
geoffwhittington: ConnectParam should probably include a 'type' field so the frontend knows if it should mask the entry. Also, the backend will need to store any confidential data securely in the db.

sdetools/connector_plugin/sample_plugin.txt
```diff
@@ -0,0 +1,41 @@
...
+                                   description="This is a password",
+                                   label="Password")
+
+        return [username, password]
+
+    @property
+    def expected_custom_attributes(self):
```
geoffwhittington: I would exclude Custom - since this term is SDE-specific, the plugin just needs attributes.

sdetools/connector_plugin/plugin.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+ConnectionParam = namedtuple("ConnectionParam", ["name", "description", "label"])
+CustomAttribute = namedtuple("CustomAttribute", ["name", "type"])
```
geoffwhittington: Any reason why a `ConnectionParam` would be any different from a `CustomAttribute`?

sdetools/connector_plugin/manager.py
geoffwhittington: Does it make sense to have more than one plugin manager running during program execution? I like the idea of using the singleton pattern - though I'm not clear how this sort of thing works in Python.

emintham: You'd use the Borg pattern in Python instead. Basically boils down to
having all instances of the class sharing the same __dict__.

```python
class MyBorg(object):
    __shared_dict__ = {}
    def __init__(self, ...):
        self.__dict__ = self.__shared_dict__
        # ...
```

Though, in this case, we can let the metaclass handle the registering and
we won't have to instantiate multiple Managers anyway.

geoffwhittington: Thinking about singleton's more - I'm not convinced its that much better than the alternative in the grand scheme of things

hanifvirani: (1)

```
manager = ProjectConnectorPluginManager(directory)
manager.get_plugins()
```
 
(2)

```
ProjectConnectorPluginManager.get_plugins(directory)
```

I feel (2) seems better than (1)

sdetools/connector_plugin/plugin.py
hanifvirani: I am thinking this will return something like {'error': 'blah'} on failure.

sdetools/connector_plugin/plugin.py
hanifvirani: No particular reason. Can change it.

sdetools/connector_plugin/sample_plugin.txt
hanifvirani: I don't think there is anything wrong with properties doing instantiations. These instantiations are specific to this property.

sdetools/connector_plugin/sample_plugin.txt
hanifvirani: Would this then also return non custom SDE attributes?

sdetools/connector_plugin/plugin.py
hanifvirani: The way we are using a CustomAttribute type in this framework, it doesn't require a description and label.

hanifvirani:
> My preference is to refer to these as configuration_params - its more general so we can stuff other things (behaviour instructions) in here as well as connection information. Of course once we do call it configuration_params we could put the expected_custom_attributes as configuration into here as well.

What are people's thoughts on this? I agree that configuration params is more general, but I am not sure about stuffing expected_custom_attributes in there. It will muddy the return structure.

hanifvirani:
> How does a plugin user know to make use of name and description?

Through documentation.

sdetools/connector_plugin/sample_plugin.txt
hanifvirani: What would be the expected valid types?

sdetools/connector_plugin/__init__.py
hanifvirani: Fixed.

sdetools/connector_plugin/sample_plugin.txt
emintham: Ahhh I missed the fact that those were namedtuples. In this case, it's fine.

sdetools/connector_plugin/plugin.py
sibartlett: SDE is currently expecting errors to be like:

```
{
  "errors": [
    "I am an error",
    "I am another error"
  ]
}
```

sibartlett:
Yeah, I also agree that "configuration params" is a better name than "connection params".

I don't think "expected_custom_attributes" should be nested under it. Unless we talking about nesting both sets of "configuration params" and "expected_custom_attributes" under another data structure - another layer might be over complicating things though?

sdetools/connector_plugin/plugin.py
geoffwhittington: My personal favourite is Jira's error response:

```
{
    "errorMessages" : [
       "This is error1",
       "This is error2"
    ],
    "errors": {
       "field1": "This really should be an integer",
       "field2": "Does not follow the expected format"
    }
}
```

sibartlett:
Named tuples were introduced in Python 2.6, but sdetools is still targeting 2.4 - we'll need to discuss an alternative.

emintham:
Good catch Simon!

`namedtuple` is written in pure Python under collections.py. We can pull that code if we do want to use it or we can just return regular tuples.

sdetools/connector_plugin/manager.py
funkaoshi: Have we solved the hardcoding the plugin directory problem yet? This can't be merged while we are assuming this stuff.

hanifvirani: Yeah, you are able to pass in the directory.

geoffwhittington: Can we leave it unset, make the consumer specify it, and throw an exception if it is not set?

hanifvirani: Right now if you don't pass a plugins_folder param, we fall back to a default one: '/docs/sde/connector-plugins/'

sdetools/connector_plugin/plugin.py
```diff
@@ -0,0 +1,96 @@
...
+        pass
+
+    @abstractmethod
+    def fetch_project_data(self, configuration, lookup_data):
+        """
+        Returns the data for the project associated with the given id.
+
+        Args:
+
+            id:            Id of the project in the external system.
```
geoffwhittington: Docs need updating to reflect the new param

sdetools/connector_plugin/exceptions.py
```diff
@@ -0,0 +1,89 @@
...
+    general errors. A dict containing a key, value pair of field names, and
+    a list of error messages can be passed for field specific errors.
+    Alternatively NON_FIELD_ERRORS key can also be used for general errors
+
+    E.g
+    ProjectConnectorException("General error message")
+
+    ProjectConnectorException(
+        ["General error message 1",
+        ["General error message 2"]
```
sibartlett: Remove extra bracket.

sdetools/connector_plugin/manager.py
```diff
@@ -0,0 +1,74 @@
...
+                    modname, ext = os.path.splitext(filename)
+
+                    if ext == ".py":
+                        filename, path, desc = imp.find_module(modname, [directory_path])
+                        plugin_modules.append((modname, filename, path, desc,))
+
+        return plugin_modules
+
+    @classmethod
+    def get_plugins(cls, plugins_folder=None):
```
sibartlett: Can we also add these methods:

```
def get_plugin(cls, plugins_folder, plugin_id)
def get_plugin(cls, plugin_path)
```

sibartlett: Actually the second one isn't so important right now - more a nice to have.

hanifvirani: added get_plugin

sdetools/connector_plugin/manager.py
funkaoshi: Should call this `DEFAULT_PLUGIN_FOLDER`. I'm also not sure we should have this. Why not make it a requirement to pass in the location of the folder?

sdetools/connector_plugin/manager.py
```diff
@@ -0,0 +1,85 @@
...
+            modname, filename, path, desc = plugin_module
+
+            if filename:
+                # loading the module auto-registers it
+                imp.load_module(modname, filename, path, desc)
+
+        return cls._plugins
+
+    @classmethod
+    def get_plugin(cls, plugin_id, plugins_folder=None):
```
funkaoshi: As above, i'd make plugin_folder required, and then have the definition be: `def get_plugins(cls, plugin_folder, plugin_id)`

sdetools/connector_plugin/plugin.py
funkaoshi: What Geoff said.

sdetools/connector_plugin/types.py
```diff
@@ -0,0 +1,42 @@
+class ConnectorParamType(object):
+    """
+    The type of configuration parameter.
+
+    STRING:         A normal text field.
+    MASKED_STRING:  A masked text field (e.g. password)
+    """
+
+    STRING = 1
```
funkaoshi: Seems weird we don't use strings here as well to match the other "type".

funkaoshi:
Small comments above. Seems close. We can wrap this up and merge.

hanifvirani:
Fixed MR comments.

PR #289, opened 2015-08-10
Project Connector plugin examples (IMDB, JIRA and Rally)

geoffwhittington:


sdetools/connector_plugin/docs/sde/connector_plugins/jira_plugin/jira_plugin.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def get_project_details(self, id, configuration):
+        try:
+            req_url = "%s/rest/api/2/project/%s" % (configuration['server'], id)
+            request = urllib2.Request(req_url)
+            base64string = base64.encodestring('%s:%s' % (configuration['user'], configuration['pass'])).replace('\n', '')
+            request.add_header("Authorization", "Basic %s" % base64string)
+            result = urllib2.urlopen(request).read()
+            return json.loads(result)
+        except:
```
funkaoshi: We shouldn't use catch all exceptions. It makes debugging harder, as it can sometimes mask errors. We should catch the exceptions we expect, and fail on the ones we don't.

sdetools/connector_plugin/docs/sde/connector_plugins/rally_plugin.py
```diff
@@ -0,0 +1,73 @@
...
+                }
+            }
+        except KeyError:
+            msg = "Make sure you are using the right project id"
+            return {
+              "errors": [msg]
+            }
+
+    def get_project_details(self, id, configuration):
+        try:
```
funkaoshi: More generally, i'd only wrap things in try/except that we know will raise an exception. Generating the req_url will not, for example.

sdetools/connector_plugin/examples/imdb_plugin.py
```diff
@@ -0,0 +1,64 @@
+import json
+from urllib import urlencode
+from urllib2 import urlopen
+
+from sdetools.connector_plugin import (BaseProjectConnectorPlugin,
+                                       CustomAttribute, ConnectorParam)
+
+
```
geoffwhittington: Extra line

geoffwhittington: Sorry never mind

sdetools/connector_plugin/examples/imdb_plugin.py
```diff
@@ -0,0 +1,64 @@
...
+    @property
+    def expected_custom_attributes(self):
+        return [
+            CustomAttribute(name='year'),
+            CustomAttribute(name='rating'),
+            CustomAttribute(name='director'),
+            CustomAttribute(name='imdbID')
+        ]
+
+    def fetch_project_data(self, configuration, lookup_data):
```
geoffwhittington: What contract do these plugins need to implement regarding exceptions - will the IMDB service not returning JSON triggering an exception be caught and gracefully dealt with by SD Elements

geoffwhittington: Or more generally, will SD Elements crash because of a plugin error

funkaoshi:
I stil think we should tidy up the exception catching in the fetch data methods.

sdetools/connector_plugin/examples/imdb_plugin.py
```diff
@@ -0,0 +1,59 @@
...
+        external_id = lookup_data.get('external_id', '')
+        param_id = 'i' if external_id.startswith('tt') else 't'
+        url = 'http://www.omdbapi.com/?' + urlencode([
+            (param_id, external_id),
+            ('plot', 'short'),
+            ('r', 'json'),
+        ])
+
+        content = urlopen(url).read()
+        movie = json.loads(content)
```
funkaoshi: Shouldn't we catch connection errrors (urlopen fails) or json.loads can't parse the data it gets back?

funkaoshi:
These seem fine +1

PR #291, opened 2015-08-26
Non modifiable attributes

hanifvirani:


sdetools/connector_plugin/plugin.py
```diff
@@ -75,6 +75,14 @@ def expected_custom_attributes(self):
         pass
 
     @abstractmethod
+    def non_modifiable_attributes(self):
```
hanifvirani: I can rename this to read_only_attributes if people feel that is better. The reason I named it this is that there is nothing being "written" on the plugin. The plugin is just dictating what attributes it doesn't want SDE to modify. So in the context of a plugin, that felt better to me.

funkaoshi: I also prefer read-only. I get what you are saying, but I think read-only is probably just as clear to people.

funkaoshi:
+1

PR #295, opened 2015-09-10
Add fix for bypassing SSL cert validation

geoffwhittington:


geoffwhittington:
This has already been merged into master

PR #296, opened 2015-09-10
Various updates

geoffwhittington:


install/setup.iss
```diff
@@ -0,0 +1,122 @@
+#define EXE_VERSION "1.0"
```
funkaoshi: I don't really understand any of this. FYI.

sdetools/sdelib/conf_mgr.py
```diff
@@ -169,7 +169,7 @@ def __getitem__(self, key):
 
     def __setitem__(self, key, val):
         if key not in self.settings:
-            raise KeyError, 'Unknown configuration item: %s' % (key)
+            return
```
funkaoshi: Any reason we drop this? Might be nice to log a warning for bad configuration items. Would catch typos and the like.

setup.py
```diff
@@ -22,7 +22,11 @@ def set_py2exe_options():
 
     for mod_name in modules.__all__:
         ext_opt['options']['py2exe']['includes'].append('sdetools.modules.%s' % mod_name)
-    
+
+    from sdetools import extlib
```
funkaoshi: Is there a circular import if this is outside the function?

install/setup.iss
geoffwhittington: That's fine. This is a functioning installer but its not perfect.

This particular macro is for updating the file version of the generated setup.exe. I included it just to be able to differentiate

funkaoshi: Cool. Yeah, I mean I think I understand what it's doing, but I don't think i can offer you any good commentary on whether it's amazing or not. It probably is.

funkaoshi:
Seems fine @geoffwhittington

setup.py
```diff
@@ -22,7 +22,11 @@ def set_py2exe_options():
```
funkaoshi: Explain what this does maybe? I don't think i'd know off hand.

PR #297, opened 2015-09-10
App scan rest

wisznie:


geoffwhittington:
What were the two (or more) issue statuses we wanted to use as blacklists?

wisznie:
Not 100% sure, but I think it was issue status, so we could blacklist issues marked as noise or passed etc. Added in that functionality.


On September 14, 2015 at 00:58:53, geoffwhittington (notifications@github.com) wrote:

What were the two (or more) issue statuses we wanted to use as blacklists?

geoffwhittington:
Merging in #346

PR #303, opened 2015-11-01
Feature/versionone

wisznie:
Initial test logic ready, just need sample response files and should be mostly good.


sdetools/sdelib/restclient.py
```diff
@@ -218,7 +218,7 @@ def call_api(self, target, method=URLRequest.GET, args=None, call_headers={}, au
             if args:
                 req_url = '%s?%s' % (req_url, urllib.urlencode(args))
         else:
-            data = self.encode_post_args(args)
+            data = self.encode_post_args(args, call_headers)
         req = URLRequest(req_url, data=data, method=method)
```
wisznie: It's a small change that shouldn't break past integrations and should prevent needing to override the method in the versionone plugin class.

PR #305, opened 2015-11-03
Allow multiple transitions to close issue

geoffwhittington:


geoffwhittington:
@ehsanf This will do the trick to support more than one transition. The impact is there is one web service call for every transition to validate that the move is correct. One work around is to stop checking and just fail.

PR #302, opened 2015-10-16
Feature/versionone

geoffwhittington:

PR #309, opened 2015-11-11
Feature/versionone

wisznie:


wisznie:
It was a case of a careless pull request. I could have sworn I had the version one branch selected. Those are the working tests. It also includes 1 line changes for mingle and hp because of that small change to RESTClient we talked about the other week, it needed to send an additional argument to work with V1's mix of xml and JSON. Rather than overriding a large chunk of code in V1 I added the additional argument in RESTClient that doesn't break anything and works fine.

wisznie:
So you may want to merge it in entirely.

geoffwhittington:
Closing since changes have been merged into #302

PR #313, opened 2015-11-26
Convert sdetools to APIv2

jnlantos:


sde_int_api.py
```diff
@@ -11,6 +11,10 @@ class InternalAPI(restclient):
     def __init__(self, conf_prefix, conf_name, config):
         pass
 
+    def get_business_units(self, **filters):
+        response = self.call_api('business-units/', args=filters)
```
funkaoshi: I wonder if the call_api command should tack on the '/' when it's called if it's missing? Something to think about in the future.

sdetools/__init__.py
```diff
@@ -7,7 +7,8 @@
     'extlib',
 ]
 
-VERSION = '4.8.4'
+VERSION = '4.9.0'
```
funkaoshi: Should probably bump this to 5.0. It feels like a big change.

sdetools/modules/sync_versionone/versionone_plugin.py
```diff
@@ -353,7 +353,7 @@ def alm_add_task(self, task):
 
         # Update the status to match task
         if (self.config['alm_standard_workflow'] and
-                (task['status'] == 'DONE' or task['status'] == 'NA')):
+                (task['status']['meaning'] == 'DONE' or task['status']['meaning'] == 'NA')):
```
funkaoshi: You could have also re-written this as `task['status']['meaning'] in ['DONE', 'NA']` as you have above.

sdetools/sdelib/testlib/sde_response_generator.py
```diff
@@ -75,197 +120,204 @@ def generate_sde_task(self, task_number=None, project_id=None, status=None, prio
...
-    def get_projects(self, target, flag, data, method):
-        if not flag:
-            projects = self.generator_get_filtered_resource('project', self.get_url_parameters(target))
+                return RESPONSE_HEADERS, ''
+        self.raise_error('405')
 
-            return RESPONSE_HEADERS, {'projects': projects}
-        else:
+    def get_projects(self, target, flag, data, method):
+        if flag:
```
funkaoshi: We should rename `flag` to `raise_401` i'd say. Should have a plan to clean this stuff up.

funkaoshi:
I say merge. I blitzed through this pretty quick because I assume it's been looked at a few times now.

sde_int_api.py
geoffwhittington: I hope this is a temporary consideration - as we should make a trailing / optional in the API

sdetools/sdelib/sdeapi.py
```diff
@@ -39,25 +39,55 @@ def post_conf_init(self):
...
+    def create_user(self, email, first_name, last_name, role=None, is_active=True, groups=None):
+        """
+        Create a user
+        """
+        args = {
+            'email': email,
+            'first_name': first_name,
+            'last_name': last_name,
+            'is_active': is_active,
+            'role': role,
```
geoffwhittington: We should remove this since we are already creating a role in the next statement

sdetools/sdelib/sdeapi.py
```diff
@@ -39,25 +39,55 @@ def post_conf_init(self):
...
+            'groups': groups
```
geoffwhittington: This should be moved out and rewritten like `role`

PR #320, opened 2015-12-10
Add support for filtering remote syncs.

funkaoshi:
In an upcoming version of SDE we will return if the connection is
inaccessible or not from within SDE. These are the connections the
command driver should run.

Also, the command driver seems like a weird name for this.


funkaoshi:
@geoffwhittington This is so untested you have no idea.

geoffwhittington:
Those are the best merges

geoffwhittington:
Command driver is called as such because it runs other commands. Orchestrator? Conductor?

sdetools/modules/command_driver.py
```diff
@@ -47,8 +47,11 @@ def in_scope(self, connection):
         """
         Check to see if a connection is in scope
         """
-        in_scope = True
+        # To start, we are only interested in connections that have been
+        # flagged as inaccessible within SD Elements ...
```
geoffwhittington: It is conceivable that users will want to run against accessible connectors too

sdetools/modules/command_driver.py
```diff
@@ -53,6 +59,12 @@ def in_scope(self, connection):
             connection_combo_id = connection['type'] + '-' + str(connection['id'])
             in_scope =  connection_combo_id in self.config['filter_connections'].split(",")
 
+        # Filter connections based on accessibility
+        if in_scope and self.config['filter_connection_access'] == 'inaccessible':
+            in_scope = connection['inaccessible'] == True
```
funkaoshi: `in_scop = in_scope and connection['inaccessible'] == True`

sdetools/modules/command_driver.py
```diff
@@ -53,6 +59,12 @@ def in_scope(self, connection):
...
+        elif in_scope and self.config['filter_connection_access'] == 'accessible':
+            in_scope = connection['inaccessible'] == False
```
funkaoshi: `in_scope = in_scope and connection['inaccessible'] == False`

funkaoshi:
This looks fine, but are we actually going to want these filtering options? Or is it just more work for the enduser?

sdetools/modules/command_driver.py
geoffwhittington: good catch thanks!

geoffwhittington:
If we want to debug a failing integration that normally runs on the server we need to be to do that. Also, for people who have to run their jobs off-site because the server is inaccessible, we default that behaviour. It may also be the case that a user may want to run all jobs regardless of whether the server is accessible or not.

geoffwhittington:
@funkaoshi If you're good with this, I'm going to squash and merge

PR #321, opened 2015-12-16
Use task-status id instead of meaning to update sde statuses

jnlantos:
The sdetools side of SDE-1922


ehsanf:
@jnlantos the only downside to it is that it makes the code harder to read (e.g. it is easier to read if status == 'DONE' than status == 'TS1'). I recommend defining a constant somewhere like:

class STATUS:
     DONE = 'TS1'
     ...

and have status = STATUS.DONE

this way you get both benefits

geoffwhittington:
Good initiative @jnlantos 

The neat thing about using the status id is at some point in the future we can start to allow people to dictate which statuses to use - much like we do with scanner report imports.

lmontopo:
+1

PR #324, opened 2016-01-12
Add support for priority field in Rally

geoffwhittington:


geoffwhittington:
Already merged in with #344

PR #325, opened 2016-01-14
Various fixes

sibartlett:
* Added requirements.txt
* Changed `#!/usr/bin/python` to `#!/usr/bin/env python`
  * this uses python on the user's path instead of a hard-coded python
  * allows virtualenv users to run `./sde.py` instead of `python ./sde.py`


funkaoshi:
+1

PR #326, opened 2016-01-16
LDAP Sync

geoffwhittington:


sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,383 @@
...
+        self.config.opts.add("ldap_method", "protocol of LDAP connection", default="LDAP")
+        self.config.opts.add("bind_dn", "DN of LDAP user")
+        self.config.opts.add("bind_password", "password of LDAP user")
+        self.config.opts.add("group_mapping", "mapping of AD/LDAP Groups to SD Elements")
+        self.config.opts.add("ldap_filter", "filter for users and groups during synchronization")
+        self.config.opts.add("deactivation", "toggle to automatically deactivate groupless users in SDE", default="False")
+        self.config.opts.add("ldap_validate_cert", "boolean to determine if TLS certificate should be validated", default="True")
+        self.config.opts.add("server_type", "LDAP or AD", default="LDAP")
+        self.config.opts.add("user_schema", "Schema", default='')
+        self.config.opts.add('sde_businessunit', "SDE Business Unit", default='', group_name="SD Elements Connector")
```
geoffwhittington: @emintham I have no idea why we we're asking for business unit, application, project ...

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,383 @@
+import ldap
```
geoffwhittington: New dependency

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,383 @@
...
+        
+        logger.info("Updated existing SD Elements users ")
+
+        for dir_user_id, dir_user in directory_users.iteritems():
+
+            # Check if user exists already in SD Elements
+            if not dir_user["exists"]:
+                groups = [grp['id'] for grp in self.translate_ldap_group_to_sde_group(dir_user["groups"])]
+
+                # Calculate first name based on full name and last name, if it is not available
```
geoffwhittington: @emintham This seems to happen - that we don't get a first name. Our API seems to require first/last name - even though we don't need it for our other LDAP SSO - where logged-in users are directed to set their first/last names on login

sdetools/modules/sync_ldap/ldap_module.py
emintham: Tested locally. Works even without those three lines.

emintham: Seems it was there only to satisfy the tests. I'll have an MR up with various fixes.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,383 @@
...
+
+            updated_user = {
+                    "email": sd_user["email"],
+                    "first_name": sd_user["first_name"],
+                    "last_name": sd_user["last_name"],
+                    "role": sd_user["role"],
+                    "is_active": True,
+                }
+
+            deactivate_user = False
```
emintham: Unused variable. I don't see deactivation mentioned in any emails you sent me. Not sure if this is part of the SOW @geoffwhittington ?

emintham: Seems it's in the ticket so I'll add this feature in a separate PR tomorrow.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,383 @@
...
+            if sd_user["email"] in directory_users:
+
+                # Track that this directory user exists in SD Elements
+                directory_users[sd_user["email"]]["exists"] = True
+
+                dir_user = directory_users[sd_user["email"]]
+
+                groups = [grp['id'] for grp in self.translate_ldap_group_to_sde_group(dir_user["groups"])]
+
+                # Deactivate user
```
geoffwhittington: @emintham This is where the user gets deactivated - I don't think we need more code. That line of code up there on line 304 is an oops and should be removed. I was playing with the code and probably left it. I think deactivate works ok with the code as-is....

setup.py
```diff
@@ -71,5 +71,6 @@ def set_py2exe_options():
         "git+https://git@github.com/sdelements/yantra.git@master#egg=yantra"
     ],
     data_files=static_files,
+    install_requires='python-ldap',
```
RouganStriker: Wouldn't this either overwrite or be ignored because install_requires is already declared above?

`
install_requires=[
         "yantra"
],
`

emintham: Good job! Think $CLIENT also hit that error on the call. But we kinda forgot about it because the virtual environment had those two dependencies anyway.

geoffwhittington:
@emin Can you fix tests, squash and merge - when you get free time?

sdetools/sdelib/sdeapi.py
```diff
@@ -44,16 +51,53 @@ def connect(self):
...
         return bool(self.me)
 
+    def call_api(self, target, method=URLRequest.GET, args=None,
+                 call_headers={}, auth_mode=None, page_size=0):
+        if not page_size:
+            return super(ExtAPI, self).call_api(
+                target, method, args, call_headers, auth_mode
+            )
+
+        # TODO: parse out args vs query params
```
funkaoshi: you can use urlparse to do all the url parsing junk, perhaps?

sdetools/sdelib/sdeapi.py
```diff
@@ -44,16 +51,53 @@ def connect(self):
...
+        if target.endswith('/'):
+            target += '?'
+        else:
+            target += '&'
+
+        results = {'results': []}
+        page_num = 1
+
+        while True:
+            current_page = target + 'page_size=%d&page=%d' % (page_size, page_num)
```
RouganStriker: I think you can add page_size and page to the args variable? It should append and url encode those parameters

sdetools/sdelib/sdeapi.py
RouganStriker: The call_api already does that if you pass the parameters to args

requirements.txt
```diff
@@ -1,2 +1,3 @@
 mock==1.3.0
 nose==1.3.7
+python-ldap
```
harry-kim: don't need a version number?

setup.py
```diff
@@ -66,7 +66,8 @@ def set_py2exe_options():
     version=sdetools.VERSION,
     url='https://github.com/sdelements/sdetools',
     install_requires=[
-        "yantra"
+        "yantra",
+        "python-ldap"
```
sibartlett: `sdetools` can't have any pypi dependencies. All libraries that `sdetools` depends on, should live in `extlib` directory.

We can remove `yantra` from this list, as it's already in `extlib`.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,835 @@
...
+        self.config = config
+        self._add_config_options()
+
+    def _add_config_options(self):
+        """Adds LDAP specific config options to the config file"""
+        self.config.opts.add("ldap_server",
+                             "endpoint of LDAP server")
+        self.config.opts.add("ldap_method",
+                             "protocol of LDAP connection",
+                             default="LDAP")
```
RouganStriker: Minor comment, if we are always converting this value to lowercase then maybe we should set the default value to the lowercase version?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,835 @@
...
+        """
+        new_user = {
+            'first_name': user.get('first_name', ''),
+            'last_name': user.get('last_name', ''),
+            'full_name': user.get('full_name', '')
+        }
+
+        # No first name in dict
+        if new_user['full_name']:
+            name_split = new_user['full_name'].split(' ')
```
RouganStriker: Can also do something like ```new_user['full_name'].rsplit(' ',1)``` to get the first and last name

setup.py
sibartlett: Usually we drop the latest version into that directory. We have only been copying the files necessary for it to run (no tests, etc).

However, it looks like python-ldap is a bit more complicated than that, it has custom native code that needs to be compiled?

@geoffwhittington any thoughts here?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,835 @@
...
+
+            return DEACTIVATED
+
+        sd_user_groups = [grp['id'] for grp in sd_user["groups"]]
+
+        profile_changed = False
+        profile_changed |= (sd_user['first_name'] != updated_user["first_name"])
+        profile_changed |= (sd_user['last_name'] != updated_user["last_name"])
+        profile_changed |= (sd_user['is_active'] != updated_user["is_active"])
+        profile_changed |= (sd_user_groups != groups)
```
harry-kim: could use any() here
https://docs.python.org/2/library/functions.html#any

sdetools/sdelib/testlib/sde_response_generator.py
```diff
@@ -11,9 +11,9 @@ def __init__(self, config, test_dir=None):
         self.default_app_id = 874
         self.default_bu_id = 1
         self.default_alm_connection = 2
-        self.bu_name = config['sde_businessunit']
-        self.app_name = config['sde_application']
-        self.project_name = config['sde_project']
+        self.bu_name = config.get('sde_businessunit', None)
```
harry-kim: don't need the None, it's redundant

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,835 @@
...
+
+    @property
+    def current_user(self):
+        if self._current_user:
+            return self._current_user
+
+        if self.sde_plugin:
+            self._current_user = self.sde_plugin.get_users(myself_only=True)
+
+        return self._current_user
```
emintham: We want to save it first so that we don't need to refetch from plugins the next time we query for it. But it could be moved to be under the second if block.

setup.py
sibartlett: We can keep it as a pypi dependency for now? But we must ensure that sdetools, still works without it for the most part. The following should work:

1) Clone sdetools repo
2) Do not pip install any dependencies
3) run `./sde.py help`

sdetools/modules/sync_ldap/ldap_module.py
geoffwhittington: I think we should default to `ldaps` here?

setup.py
emintham: It doesn't.... :( It'll try to import the sync_ldap module (which imports ldap) and fails.

geoffwhittington: can we import ldap at the last minute instead? Or do we prefer to import ldap earlier....

emintham: Moved all imports to be inside methods.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,841 @@
...
+        dn_regex = re.compile("dc=", re.I)
+
+        basedn = (self.normalize_dc(chunk)
+                  for chunk in dn_split
+                  if dn_regex.match(chunk))
+
+        return ','.join(basedn)
+
+    def initialize(self):
+        import ldap
```
emintham: Yeh. There were some difficulties vendoring the entire python-ldap lib into extlib so we're deferring the imports until actually needed. See @sibartlett's comment  thread above.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,841 @@
...
+        else:
+            self.config['user_schema'] = DEFAULT_SCHEMA[self.server_type]
+
+        self.user_schema = self.config['user_schema']
+
+        self.connection = ldap.initialize(
+            "%s://%s" % (self.ldap_method, self.config["ldap_server"])
+        )
+
+        # Prevents obscene query times
```
emintham: This was a comment left by the contractor in the original code for LDAP... I think I saw it mentioned in someone's blog as well.. I'll try to find a reference

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,841 @@
...
+        return self.parse_groups(ldap_groups, self.config['group_mapping'])
+
+
+class LDAPConnector(object):
+    """
+    Connector for sync-ing LDAP/AD servers with SDE.
+
+    1. Create new users found on LDAP but not SDE.
+    2. Deactivate users found on SDE but not LDAP.
+    3. Deactivate users without groups.
```
emintham: I think @geoffwhittington added the code for requirement 3 specifically? Maybe it was for $CLIENT?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,841 @@
...
+
+
+logger = log_mgr.mods.add_mod(__name__)
+
+
+class LDAPConnectorError(Error):
+    pass
+
+
+DEFAULT_SCHEMA = {
```
RouganStriker: Is there a reason for the 2 schemas? They seem to be structured the same way

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,841 @@
...
+                error_msg = "Unable to start TLS on connection: %s"
+                raise LDAPConnectorError(error_msg % e.message["desc"])
+
+        log_message = "%s connection established to %s"
+        logger.info(log_message, self.ldap_method, self.config["ldap_server"])
+
+        try:
+            self.connection.simple_bind_s(self.config["bind_dn"], self.config["bind_password"])
+        except ldap.INVALID_CREDENTIALS:
+            raise LDAPConnectorError("Invalid credentials")
```
RouganStriker: Can we expand on this error message and ask the user to check their bind_dn and bind_password configurations?

RouganStriker:
Should we be verifying that there are no duplicate emails among the user data pulled from LDAP? Or is that out of scope/unneeded

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,841 @@
...
+    def fetch_user_data(self):
+        """
+        returns a 5-tuple containing:
+        (SDE users, LDAP users, users in SDE and LDAP,
+         users in LDAP but not SDE, users in SDE but not LDAP)
+        """
+        sd_users = dict(
+            (user['email'], user)
+            for user in self.sde_plugin.get_users()['results']
+            if (self.current_user['id'] != user['id'] and
```
RouganStriker: This filters out all super users and the current user from the list of SDE users right? If a user with their email exists in the LDAP server then the sync will attempt to create them (but will not succeed)

sdetools/modules/sync_ldap/ldap_settings.conf.example
```diff
@@ -0,0 +1,50 @@
...
+bind_dn = cn=Administrator,cn=Users,dc=example,dc=org
+
+#LDAP credentials
+bind_password = password
+
+#Mapping of LDAP group names to SD Elements group names
+group_mapping = {"LDAPGroup" : "SDEGroup", "LDAPGroup3" : "SDEGroup3"}
+
+#Filter for which groups and users should be synced
+ldap_filter = {"groups": ["Group", "Group3"], "users": ["user@example.org", "user2@example.org"]}
```
RouganStriker: Can we update the example to be "LDAPGroup", "LDAPGroup3" since it uses the ldap group names?

sdetools/modules/sync_ldap/ldap_module.py
emintham: Main difference is gn vs givenName for LDAP vs AD. This is the mapping from SDE fields to LDAP fields. I could have just one schema and do an if check in the code?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,841 @@
...
+
+            paging_control.cookie = cookie
+
+        logger.info('Pulled %s pages of ldap queries for a total of %s results',
+                    pages, len(results))
+
+        return results
+
+    def get_user_params(self):
+        params = [
```
RouganStriker: Can also do params = self.user_schema.values()

sdetools/modules/sync_ldap/ldap_module.py
emintham: Good catch!

sdetools/modules/sync_ldap/ldap_module.py
emintham: According to @geoffwhittington it was so that we could exclusively rely on LDAP.

cc @RouganStriker re: comment about deactivation below.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Added comment

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -16,25 +16,91 @@ class LDAPConnectorError(Error):
...
+            1: 'Windows Server 2003 Interim',
+            2: 'Windows Server 2003',
+            3: 'Windows Server 2008',
+            4: 'Windows Server 2008'
+        }
+
+        return functionality_mapping.get(num, '')
+
+    def __str__(self):
+        log_message = ''
```
RouganStriker: Can we also include the server_type in this log message?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,928 @@
...
+            logger.debug(error_message, sd_user["email"])
+
+            updated_user.update(is_active=False)
+
+            return DEACTIVATED
+
+        sd_user_groups = [grp['id'] for grp in sd_user["groups"]]
+
+        profile_changed = any([
+            sd_user['first_name'] != updated_user["first_name"],
```
RouganStriker: The values in updated_user are not unicode strings while the ones in sd_user are and this is generating warning messages in the console. We can convert everything to unicode to clean up the logs: ```updated_user["first_name"].decode('utf-8')```

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,928 @@
...
+            sd_user['last_name'] != updated_user["last_name"],
+            sd_user['is_active'] != updated_user["is_active"],
+            sd_user_groups != groups
```
RouganStriker: Should convert to sets to because unordered lists can return false in comparison even if they contain the same elements

sdetools/modules/sync_ldap/ldap_module.py
geoffwhittington: Should we make this mapping configurable? If so, maybe we can do it in another release

sdetools/modules/sync_ldap/ldap_module.py
geoffwhittington: nvm, it is ha!

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -0,0 +1,928 @@
...
+            )
+
+        if not self.config['bind_dn']:
+            raise LDAPConnectorError("Missing value for bind_dn")
+
+        # Setup the schema if needed
+        self.user_schema = DEFAULT_SCHEMA
+        if self.config['user_schema']:
+            self.config.process_json_dict('user_schema')
+            self.user_schema.update(self.config['user_schema'])
```
RouganStriker: I think we should apply the gn => givenName change here so the LDAPConnector methods have access to the correct user_schema

RouganStriker:
Is it possible to change a user's full name in the LDAP server? I can change the first name and last name but that doesn't seem to update the full name. This probably isn't  a big issue since names rarely change but at the moment I think we aren't ever updating a user's name.

Asides from the comments this PR is looking good, nice job on the auto AD/LDAP detection thing!

sdetools/modules/sync_ldap/ldap_module.py
emintham: The server_type isn't available at this point after the server_type inferring change. It's only available after the .connect() call. I looked at the call sites of the user_schema in LDAPConnector and the user_schema is only required after LDAP connects and pulls data so it shouldn't be a problem.

RouganStriker: I had an issue in filter_users where it couldn't find the user's first_name because the mapping still mapped it to gn instead of givenName

emintham: Oh right. I see what you mean now. I've made the change to the end of .connect() after server_type is available.

RouganStriker:
+1, looks good to me

PR #328, opened 2016-02-01
Hotfix/include parent jira

funkaoshi:
This code isn't in master?


geoffwhittington:
What's the status of this MR?

funkaoshi:
This doesn't even have the change.

PR #330, opened 2016-02-01
Fix/sync ldap

emintham:


sdetools/modules/sync_ldap/tests/ldap_response_generator.py
```diff
@@ -17,24 +15,46 @@ def connect():
...
+                     'CN=David Hilbert,CN=Users,DC=labs,DC=sdelements,DC=com',
+                     'CN=john neumann,CN=Users,DC=labs,DC=sdelements,DC=com'
+                  ],
+                  'name': ['mathematicians']}),
+                ('CN=scientists,CN=Users,DC=labs,DC=sdelements,DC=com',
+                 {'member': [
+                    'CN=john smith,CN=Users,DC=labs,DC=sdelements,DC=com',
+                    'CN=robert openheimer,CN=Users,DC=labs,DC=sdelements,DC=com'
+                  ],
+                  'name': ['scientists']})
```
lmontopo: It seems to me like the value of 'name' here is always going to be a list of one element.  why not just return the string without the list container?  Then, in other parts of the code we wouldn't have to index into name? Maybe i'm misunderstanding something?

lmontopo: Oups i'm looking at the test.  Doh i'm just being silly here.  That makes things a lot clearer and also makes more sense as to why things are hardcoded! haha i was so confused.  sorry!

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -110,70 +143,125 @@ def connect(self):
...
+                params = [
+                    self.user_schema['email'],
+                    self.user_schema['full_name'],
+                    self.user_schema['last_name'],
+                    self.user_schema['first_name'],
+                ]
+                user = self.connection.search_s(self.basedn,
+                                                ldap.SCOPE_SUBTREE,
+                                                uid,
+                                                params)
```
lmontopo: Nice clean up here!

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -195,39 +286,47 @@ def validate_configuration(self):
...
                 # Skip users that do not have an email address
-                if self.config['user_schema']['email'] not in user:
+                if email_key not in user:
                     continue
 
-                # Skip disabled users (userAccountControl: 512 -> enabled, = 514 -> disabled)
-                if "userAccountControl" in user and user["userAccountControl"][0] != "512":
+                # Skip disabled users
+                # disabled users have bit flag set to 1 at the 2's position
+                # see (https://support.microsoft.com/en-us/kb/305144)
```
lmontopo: Nice comments here.  I would have been super confused without them.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -265,43 +364,91 @@ def translate_ldap_group_to_sde_group(self, groups):
             translation.append(self.groups[sde_group])
 
         return translation
-            
+
+    def deactivate_user(self, sd_user, groups):
+        if sd_user['is_active'] and not sd_user['groups'] and not groups:
+            error_message = ("SD Elements user '%s' belongs to no groups and "
+                             "will be disabled")
```
lmontopo: Is there an extra " here?

sdetools/modules/sync_ldap/ldap_module.py
emintham: Yeh I had to do some research to find out why it wasn't working and fixed that-- don't want to have to spend the same amount of time to figure it out again in the future.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Thanks.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Where? I'm seeing
```
+            error_message = ("SD Elements user '%s' belongs to no groups and "
+                             "will be disabled")
```

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -66,20 +90,24 @@ def initialize(self):
...
-        if self.config['ldap_method'] not in ['ldap', 'ldaps']:
-            raise LDAPConnectorError("Invalid value for ldap_method: %s" % self.config['ldap_method'])
+        if self.config['ldap_method'].lower() not in ['ldap', 'ldaps']:
+            raise LDAPConnectorError(
+                "Invalid value for ldap_method: %s" % self.config['ldap_method']
+            )
 
-        if self.config['server_type'] not in ['LDAP', 'AD']:
-            raise LDAPConnectorError("Invalid value for server_type: %s" % self.config['server_type'])
+        if self.server_type not in ['ldap', 'ad']:
```
lmontopo: Not a big deal, but just a suggestion: You created
``` 
@property
def server_type(self):
    return self.config['server_type'].lower()
```
to transform `self.config['server_type']` to a lowercase property, but you didn't do this for `self.config['ldap_method']`.  I think it would be nice to be consistent:  i.e. Either create the property for both variables or for neither variable.

sdetools/modules/sync_ldap/ldap_module.py
lmontopo: you're right there isn't an extra one.  I just got confused cause it was split up onto two lines and that threw me off for some reason.  But you're completely right - this is not a problem.

geoffwhittington:
@emintham We don't have a lot of exception handling (neither does the branch I initiatied)- what should happen if a create_user call fails - do we let the whole sync die or can we catch it, log it, and then move on?

geoffwhittington:
Taking a step back ... I wonder if in the future we will want to refactor this so the sync_ldap subclasses some new parent class - a class which deals with user syncing - much like we have for alm_plugin_base.py. Maybe that's a conversation for when we have something other than LDAP/AD to sync with...

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -265,43 +364,91 @@ def translate_ldap_group_to_sde_group(self, groups):
...
+        Calculate first name based on full name and last name if it is not
+        available.
+        """
+        new_user = {
+            'first_name': user['first_name'] if 'first_name' in user else '',
+            'last_name': user['last_name'] if 'last_name' in user else '',
+            'full_name': user['full_name'] if 'full_name' in user else ''
+        }
+
+        if 'first_name' in new_user:
```
lmontopo: wait, based on line 382, won't this line always be true?

emintham:
@geoffwhittington Right now it'll throw an APIError should the create_user fail and the sync will stop at that point (all previous successes will still be successes). Would you prefer to catch and log failures? A concern there would be that the logging is very verbose right now and failed ones might get lost/ignored in the deluge-- right now, assuming whatever is causing the exception is fixable by the user, re-running the command again will not create bad data.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Good catch. Fixed.

sdetools/modules/sync_ldap/ldap_module.py
emintham: True. Done.

sdetools/sdelib/testlib/sde_response_generator.py
```diff
@@ -141,33 +199,35 @@ def get_users(self, target, flag, data, method, myself_only=False):
...
                 }
         else:
             self.raise_error('400')
 
+    def update_user(self, target, flag, data, method):
+        if flag:
+            self.raise_error('401')
+
+        if method == 'PUT':
+            user_id = int(target.split('/')[-2])
```
funkaoshi: Maybe comment on what target looks like?

sdetools/modules/sync_ldap/tests/ldap_module_test.py
```diff
@@ -72,40 +107,43 @@ def init_connector(self):
...
-        returnedLDAPGroups = self.connector.queryGroupsLDAP()
-        self.assertEqual(returnedLDAPGroups, expectedLDAPGroups, "Groups returned do not match groups expected")
+        returnedLDAPGroups = self.connector.query_groups_ldap()
+        self.assertEqual(returnedLDAPGroups,
+                         expectedLDAPGroups,
+                         "Groups returned do not match groups expected")
         self.connector.connection.unbind()
 
-    def testSynchronizeNewUsers(self):
+    def test_synchronize_new_member(self):
```
funkaoshi: The test is that no errors are raised or something?

emintham: It doesn't test anything.... It just does stuff using the some garbage mock data passed in... The mock data goes through our mock LDAP response generator and is just returned. Which is why I'm working on refactoring the synchronize function to be able to test this function. Apart from the assertTrue tests modified above, all the 'tests' don't really test anything.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -341,22 +473,31 @@ def synchronize(self):
                 profile_unchanged &= (sd_user['last_name'] == updated_user["last_name"])
                 profile_unchanged &= (sd_user['is_active'] == updated_user["is_active"])
                 profile_unchanged &= (sd_user_groups == groups)
- 
-                if not profile_unchanged:
-                    self.sde_plugin.update_user(sd_user["id"], email=sd_user["email"], first_name=updated_user["first_name"],
-                                                last_name=updated_user["last_name"], is_active=updated_user['is_active'], groups=groups)
 
+                if not profile_unchanged:
```
funkaoshi: The double negative here is terrible. Hah.

emintham: Agreed. Removed it in my current WIP change.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -110,70 +143,125 @@ def connect(self):
...
+        dirty_list = self._get_groups_from_ldap()
+        logger.info("Querying LDAP Server for users")
+
         groups = {}
 
-        for listItem in dirtyList:
-            items = re.split("(ou=\w*)", listItem[0])
-            groupName = items[1][3:]
+        for list_item in dirty_list:
+            items = re.split("([oO][uU]=\w*)", list_item[0])
```
funkaoshi: Might want to explain the regex?

funkaoshi:
Seems good? Feel like I can't really comment without knowing more about LDAP. Hah.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -27,36 +32,55 @@ class LDAPConnectorError(Error):
...
+        if self.config["base_dn"]:
+            return self.config["base_dn"]
+
+        dn_split = re.split(",", self.config["bind_dn"])
+
         basedn = ""
-        for item in DNsplit:
-            if re.match("dc=", item):
+        for item in dn_split:
+            if re.match("[dD][cC]=", item):
```
RouganStriker: I'm not sure if this loop will run a lot but if it does it might be better to store the compiled regex object. Also we can use the ignore case flag `re.I` instead of including both upper and lowercase c and d.

```
dn_regex = re.compile("dc=",re.I)
dn_regex.match(item)
```

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -239,18 +338,18 @@ def filter_users(self, groups, filters):
...
-                        if self.config['user_schema'][field] in user:
-                            dir_user[field] = user[self.config['user_schema'][field]][0]
-                        
+                    for field_key, field in self.user_schema.iteritems():
+                        if field in user:
+                            dir_user[field_key] = user[field][0]
+
                     users[email] = dir_user
                 else:
                     pass
```
RouganStriker: Do we need this pass?

sdetools/sdelib/testlib/sde_response_generator.py
```diff
@@ -10,9 +10,9 @@ def __init__(self, config, test_dir=None):
         self.default_project_id = 1296
         self.default_app_id = 874
         self.default_bu_id = 1
-        self.bu_name = config['sde_businessunit']
-        self.app_name = config['sde_application']
-        self.project_name = config['sde_project']
+        self.bu_name = config['sde_businessunit'] if 'sde_businessunit' in config else None
+        self.app_name = config['sde_application'] if 'sde_application' in config else None
+        self.project_name = config['sde_project'] if 'sde_project' in config else None
```
RouganStriker: We can shorten this to something like `config.get('sde_project')`

sdetools/modules/sync_ldap/ldap_module.py
emintham: Good feedback. Didn't know about that one! Thanks.

sdetools/sdelib/testlib/sde_response_generator.py
emintham: True. Good stuff.

geoffwhittington:
Sounds like we are good functionality-wise with this feature?

We should fix these tests, merge, and squash.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -26,38 +31,85 @@ class LDAPConnectorError(Error):
...
+    """
+    Connector for sync-ing LDAP/AD servers with SDE.
+
+    1. Create new users found on LDAP but not SDE.
+    2. Deactivate users found on SDE but not LDAP.
+    3. Deactivate users without groups.
+    4. Ensure equality between users that are found on both.
+    """
+
+    ou_regex = re.compile('(ou=\w*)', re.I)
```
funkaoshi: What's an OU?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -239,145 +406,242 @@ def filter_users(self, groups, filters):
...
+        sd_users, directory_users, intersection, untracked_users, tracked_users = self.fetch_user_data()
 
-                new_user_count += 1
-                self.sde_plugin.create_user(email=dir_user_id, first_name=dir_user["first_name"], 
-                                            last_name=dir_user["last_name"], groups=groups)
+        deactivated_user_count, updated_user_count = self.sync_intersection(intersection, sd_users, directory_users)
+        deactivated_user_count += self.deactivate_stale_users(tracked_users, sd_users)
+        new_user_count = self.sync_new_users_from_ldap(untracked_users, directory_users)
 
+        logger.info("Updated %d SD Elements users", updated_user_count)
```
geoffwhittington: @ehsanf You may be interested in the output of this, perhaps

sdetools/modules/sync_ldap/ldap_module.py
emintham: Organizational Unit

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -110,127 +174,230 @@ def connect(self):
...
+
+        if self.server_type == 'ad':
+            params.append('userAccountControl')
+
+        return params
+
+    def get_groups_from_ldap(self):
+        logger.info("Querying LDAP Server for groups")
+
+        if self.server_type == "ad":
```
funkaoshi: This might read easier if you assign things to variables and just have one call to search:

```
if self.server_type = "ad":
    thing1 = "(&(cn=*)(objectClass=group))"
    thing2 = ["name", "member"]
else:
    thing1 = "(ou=*)"
    thing2 = ["uniqueMember"]

return self.connection.search_s(self.basedn, ldap.SCOPE_SUBTREE, thing1, thing2)
```

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -110,127 +174,230 @@ def connect(self):
...
+            info = ldap_tuple[1]
+
+            # empty group
+            if 'member' not in info:
+                return ('', [])
+
+            group_name = info['name'][0]
+            members = info.get('member', [])
+
+            return (info['name'][0], members)
```
funkaoshi: You aren't using `group_name` here.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -110,127 +174,230 @@ def connect(self):
...
+        else:
+            items = self.ou_regex.split(ldap_tuple[0])
+
+            group_name = items[1][3:]
+            members = ldap_tuple[1]["uniqueMember"]
+
+            return (group_name, members)
```
funkaoshi: Can pull this out and use it for both if statements.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -239,145 +406,242 @@ def filter_users(self, groups, filters):
...
+        Ensure first and last names are present based on what else is available.
+        """
+        new_user = {
+            'first_name': user.get('first_name', ''),
+            'last_name': user.get('last_name', ''),
+            'full_name': user.get('full_name', '')
+        }
+
+        # No first name in dict
+        if new_user['full_name']:
```
funkaoshi: If they have a full name they won't have a first name?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -239,145 +406,242 @@ def filter_users(self, groups, filters):
...
+            name_split = new_user['full_name'].split(' ')
```
funkaoshi: If you name is "Dita Von Tesse" you'd want the last name to be "Von Tesse" and the first name to be "Dita". Names are complicated. I think it'll be hard to come up with a good heuristic here. I guess people can fix in SDE profile page.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -239,145 +406,242 @@ def filter_users(self, groups, filters):
...
+    def fetch_user_data(self):
+        """
+        returns a 5-tuple containing:
+        (SDE users, LDAP users, users in SDE and LDAP,
+         users in LDAP but not SDE, users in SDE but not LDAP)
+        """
+        sd_users = dict(
+            (user['email'], user)
+            for user in self.sde_plugin.get_users()['results']
+            if (self.current_user and self.current_user['id'] != user['id'] and
```
funkaoshi: if there is no current user we return an empty list? How would we end up in that scenario?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -239,145 +406,242 @@ def filter_users(self, groups, filters):
...
+
+        # Deactivate user
+        deactivation = self.config["deactivation"]
+        user_active = sd_user['is_active']
+        without_groups = not sd_user['groups'] and not sde_groups
+
+        if (deactivation and user_active and without_groups):
+            error_message = ("SD Elements user '%s' belongs to no groups and "
+                             "will be disabled")
+            logger.debug(error_message % sd_user["email"])
```
funkaoshi: Apparently you can do this with the logger, and it's better somehow: `logger.deubg(error_message, sd_user["email"])`. Oh, it won't do the interpolation if it doesn't need to log anything. I guess that could be a win if we were writing bots to trade stocks.

sdetools/modules/sync_ldap/tests/ldap_module_test.py
```diff
@@ -72,40 +124,306 @@ def init_connector(self):
...
+            'g2': [4, 5, 6]
+        }
+        mapping = {'mathematicians': 'g1', 'scientists': 'g2'}
+        ldap_groups = ['mathematicians', 'programmers']
+        expected = [[1, 2, 3]]
+
+        actual = self.connector.translate_ldap_group_to_sde_group(ldap_groups, mapping)
+        self.assertEqual(actual, expected)
+
+    def test_calculate_name(self):
```
funkaoshi: Could probably split this up ... if we wanted.

sdetools/modules/sync_ldap/tests/ldap_module_test.py
```diff
@@ -72,40 +124,306 @@ def init_connector(self):
...
+                'full_name': DIRECTORY[user]['cn'][0],
+                'groups': ['scientists'],
+                'last_name': DIRECTORY[user]['sn'][0],
+                'email': DIRECTORY[user]['mail'][0],
+                'exists': False
+            }
+        }
+        self.assertEqual(actual, expected)
+
+    def _seed(self, query, expected, params=None, scope=ldap.SCOPE_SUBTREE):
```
funkaoshi: Sorry, don't follow what this is doing?

funkaoshi:
This is looking really good. Much nicer.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Done

sdetools/modules/sync_ldap/ldap_module.py
emintham: Good catch. I extracted `info['name'][0]` to a variable but forgot to use it. lol

sdetools/modules/sync_ldap/ldap_module.py
emintham: Done

sdetools/modules/sync_ldap/ldap_module.py
emintham: Good catch. The old code didn't pull the first name LDAP which is a bug. Added a config so that it now pulls it from LDAP.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Overly defensive check to make user current_user exists before dereferencing 'id'. Only way self.current_user does not exist is if sde_plugin was somehow not available-- in which case, the ldap probably should fail too. I'll remove this.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Fixed all logging messages.

sdetools/modules/sync_ldap/tests/ldap_module_test.py
emintham: This basically tells the mock ldap server: "if you see <query> with <params> reply with <expected>".

funkaoshi:
If you are feeling good about things in your heart then merge!

PR #331, opened 2016-02-02
Add task tags as labels

geoffwhittington:


geoffwhittington:
@emintham can you review and merge if it is kosher?

PR #332, opened 2016-02-09
Fixes issues with urlencode_str

JeffBain:
* We make all uses of urlencode_str unicode-friendly, which should fix issues
      with unicode in tasks etc. that our original fixes didn't catch

Refs SDE-1795


setup.py
```diff
@@ -65,7 +65,8 @@ def set_py2exe_options():
     version=sdetools.VERSION,
     url='https://github.com/sdelements/sdetools',
     install_requires=[
-        "yantra"
+        "yantra",
+        "six",
```
emintham: where is six used?

sdetools/sdelib/restclient.py
```diff
@@ -217,10 +219,17 @@ def generate_token_auth_header(self):
...
         """
-        encoded_args = self.encode_dict_utf8(args)
-        return urllib.urlencode(encoded_args)
+        if isinstance(args, basestring):
+            return urlencode_str(args)
+        elif isinstance(args, dict):
+            encoded_args = self.encode_dict_utf8(args)
+            return urllib.urlencode(encoded_args)
+        raise TypeError("Unexpected Data type, expects dict or string type")
+
```
lmontopo: remove extra blank line.

sdetools/sdelib/restclient.py
```diff
@@ -23,6 +23,11 @@
 
 DEFAULT_API_TOKEN_HEADER_NAME = "Authorization"
 
+# 2/3 string compatibility checks
+try:
+    basestring
+except NameError:
+    basestring = str
```
lmontopo: Perhaps `Python 2/3 compatibility checks` would be a clearer comment.  As someone who didn't know that `basestring` was no longer a type in Python 3, I didn't understand what this comment meant at first.   In hindsight it seems obvious, but perhaps adding the word 'Python' would make it more explicit.

lmontopo:
A couple very minor comments, but otherwise this looks good to me.

setup.py
JeffBain: I dropped it because it ended up feeling silly importing it just for a string check. Forgot to remove it from this file, so I'll do that.

sdetools/sdelib/restclient.py
JeffBain: Fair enough!

PR #333, opened 2016-02-17
[WIP] Jira: Update description/remote links only on each sync

geoffwhittington:


geoffwhittington:
@funkaoshi this seems to do the trick - just need to make this work against 3.4 release code

PR #334, opened 2016-02-19
Adds Paged LDAP Querying

emintham:
- Also adds mocking for search_ext (not supported by mockldap)
- Search is now async


geoffwhittington:
What is the impact of processing the sync while visiting each page of results?

emintham:
@geoffwhittington 
If the user kept the default configs and did not manually override the page_size to something silly, there is only negligible impact. This change enables the sync to work for queries whose number of results exceed the built-in page_size on ActiveDirectory.

To make this more concrete, suppose a query for X returns 500 results. Since our page_size default is 1000, it would return 1 page whether or not we paged our results. So, there is only slight overhead to check whether there is a "next page" of results (there isn't) so we are done. However, if someone changed the page_size to, say, 10. it would then take 50 trips to collect the 500 results. In this case, there would be an impact.

The main problem this tries to solve is when a query would return > 1000 results. The builtin limit in ActiveDirectory would prevent it from returning the query and would just error out without returning anything. In this case, the only recourse we have (on our end) is to page the results so that each page has up to 1000 results.

geoffwhittington:
If there are 20k employees (for instance) the sync would store these all in memory and then attempt to sync each? Do you think it would be possible or even worthwhile to sync the pages of results as they are retrieved?

May not work if the ldap connection times out during paged syncing or if the savings (memory) are low.

emintham:
Yes, it would store everything in memory and then attempt to sync each.

It's not completely possible to do what you're suggesting because we are first retrieving all `{group: [members]}`, filtering and creating a `{member: [groups]}` and then syncing each member-- so all the data has to be present to construct the result. We could change the approach.. but it would be like rewriting the whole thing from scratch.

geoffwhittington:
Sounds fair - I forgot how the sync was working again.... thanks!

lmontopo:
This looks good to me @emintham .  +1.

PR #336, opened 2016-02-23
Feature/timestamp alm sync

RouganStriker:
Modified how JIRA's get_task api URL is generated. Did some unicode testing and found no issues with the new changes. @JeffBain can you take a look at the changes in that section?


requirements.txt
```diff
@@ -1,4 +1,4 @@
 nose==1.3.7
-
+python-dateutil
```
funkaoshi: Should pin version. You can just user the latest as of right now.

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -173,19 +173,24 @@ def get_task(self, task, task_id):
 
             if isinstance(field_value, list):
                 for list_item in field_value:
-                    condition = '%s=\'%s\'' % (field_clause, self.urlencode(list_item))
+                    condition = '%s="%s"' % (field_clause, list_item)
                     task_lookup.append(condition)
             else:
                 if field_meta['schema']['type'] == 'string':
-                    condition = '%s~"\\"%s\\""' % (field_clause, self.urlencode(field_value))
```
geoffwhittington: is `urlencode` no longer needed?

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -173,19 +173,24 @@ def get_task(self, task, task_id):
...
+                    condition = '%s~"%s"' % (field_clause, field_value)
                 else:
-                    condition = '%s=\'%s\'' % (field_clause, self.urlencode(field_value))
```
geoffwhittington: is `urlencode` no longer needed?

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -678,6 +689,23 @@ def transform_task(self, task):
         task['alm_fixed_title'] = self.get_alm_task_title(self.config, task, fixed=True)
         return task
 
+    def _determine_latest(self, alm_task, sde_task):
+        """
+        Used in synchronize() for the timestamp conflict policy to determine which
+        task status takes precedence. Extracted from synchronize() so it is easier
+        to test.
+        """
+        if 'timestamp' in sde_task:
```
geoffwhittington: should this `updated`? Also, shouldn't we look for `status_updated` first and then fail over to `updated`?

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -158,7 +158,7 @@ def get_subtask_issue_types(self):
     def get_task(self, task, task_id):
 
         task_lookup = []
-        task_lookup.append('summary~"\\"%s\\""' % self.urlencode(task['alm_fixed_title']))
+        task_lookup.append('summary~"%s"' % task['alm_fixed_title']) #self.urlencode(task['alm_fixed_title']))
```
geoffwhittington: Did you mean to delete the comment `#self.urlencode(task['alm_fixed_title']))` ?

geoffwhittington:
I saw there's a new dependency `python-dateutil` - could you have made use of https://github.com/sdelements/sdetools/blob/master/sdetools/extlib/iso8601/iso8601.py at all?

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -678,6 +689,23 @@ def transform_task(self, task):
...
+            sde_time = datetime.utcfromtimestamp(sde_task['timestamp'])
```
geoffwhittington: `updated` ?

sdetools/alm_integration/alm_plugin_base.py
RouganStriker: In api v1 the field was called timestamp. If we no longer support v1 then we can remove this conditional.

Regarding the second point, I thought the reason for creating the status_updated field was to use that instead of the updated field. Do we want to keep the "legacy" behavior of comparing last updated timestamps between SDE and ALM tasks?

sdetools/modules/sync_jira/jira_rest.py
RouganStriker: Yes - I forgot to remove that comment

sdetools/alm_integration/alm_plugin_base.py
geoffwhittington: In this version of sdetools we've switched everything over to APIv2.

I thought this code was trying to handle the case where `status_updated` might not be on the artifact. If its always going to be on the project task then we can probably strike this logic.

sdetools/modules/sync_jira/jira_rest.py
RouganStriker: It is no longer needed in this function. Before we were generating the entire URL including query parameters here. Now we pass along a dictionary of those query parameters to call_api() and it will handle the url encoding of the query parameters there.

RouganStriker:
Ah - I believe the functions in that lib would work as well. I will give it a try.

sdetools/alm_integration/alm_plugin_base.py
RouganStriker: I'll remove the api v1 conditional then.

After the SDE side of changes are merged in, the status_updated field should always be available in the json response when querying Tasks.

geoffwhittington: Sweet, sounds excellent

sdetools/modules/sync_jira/jira_rest.py
geoffwhittington: +1

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -209,12 +214,29 @@ def get_task(self, task, task_id):
...
+
+        if 'changelog' in jtask:
+            # Use the last status change date as the timestamp.
+            timestamp = jtask['fields']['created']
+            for history in reversed(jtask['changelog']['histories']):
+                for item in history['items']:
+                    if item['field'] == 'status':
+                        timestamp = history.get('created')
+                        break
+                else:
```
JeffBain: Oh man, an actual case of for ... else in the wild. Looks pretty good here, though it took me a second to follow what was up.

JeffBain:
This is looking pretty good to me.

geoffwhittington:
When you're all done can you squash and merge?

RouganStriker:
What do you mean by squash and merge?

sdetools/modules/sync_jira/tests/jira_response_generator.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, config, test_dir=None):
             '/rest/api/2/project/%s/versions' % self.project_key: 'get_project_versions',
             '/rest/api/2/issue/createmeta': 'get_create_meta',
             '/rest/api/2/issuetype': 'get_issue_types',
-            '/rest/api/2/search\?jql=project%%3D\'%s\'%%20AND%%20summary~.*' % self.project_key: 'get_issue',
+            '/rest/api/2/search?.*jql=project.*summary.*'.format(self.project_key): 'get_issue',
```
geoffwhittington: will this work < python 2.7?

geoffwhittington: Nevermind - travis tests 2.6

PR #338, opened 2016-02-24
Fix updating task status and creating new tasks with alm

lmontopo:
* Update alm_update_task_status methods to look at SDE's task status meaning instead of ID, and update alm status's accordingly
    * Update alm_add_task method to create tasks with status based on SDE task status meaning instead of ID
    * Fix erroneous log messages:  Previously had been incorrectly stating the status that tasks were updated to if SDE was being updated based on alm's info.
    * All mock tests now working
    * Fix auth headers for pivotal tracker - previously had repeated the word 'Token' in the header.

Refs: #SDE-2120


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -750,11 +750,12 @@ def synchronize(self):
                         continue
 
                     # Exists in both SDE & ALM
-                    if not self.status_match(alm_task.get_status(), task['status']['id']):
+                    if not self.status_match(alm_task.get_status(), task['status']['meaning']):
```
RouganStriker: IRC alm_task.get_status() returns the status ID so like TS1, TS2, ..., so this comparison will never be true. We need to either change what the ALM task's get_status() returns or keep the old status ID comparison here.

lmontopo: Ah, good point.  There was a similar case in the trac_plugin where I actually took out this check (and I think things still work fine without the check, but I'm going to make sure).  I'm going to look over this again and see what I can come up with. Thanks for the catch!

lmontopo: Wait, Looking over this again I think I disagree with your comment @RouganStriker .  I have updated the status_match method to make it so that it accounts for me inputting the meaning instead of the id.  Am I missing something?

RouganStriker: Yea you're right nvm

PR #339, opened 2016-02-24
Update all alms to update status based on meaning

lmontopo:
* Use global lists to hold all statuses (including custom statuses) that have a particular meaning
   * Check that the task status in question is in the appropriate list, rather than checking individual ids.
   * Fix erroneous header token in pivotal tracker (previously the header label 'Token' had appeared twice)
   * Fix erroneous log messaging when task status in SDE is updated.

Refs: #SDE-2120


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -421,8 +421,22 @@ def sde_connect(self):
...
+            if status['meaning'] == 'DONE':
+                done_list.append(status['id'])
+            elif status['meaning'] == 'NA':
+                na_list.append(status['id'])
+            else:
+                todo_list.append(status['id'])
+
+        setattr(STATUS, 'DONE_LIST', done_list)
+        setattr(STATUS, 'TODO_LIST', todo_list)
+        setattr(STATUS, 'NA_LIST', na_list)
```
RouganStriker: If we are only performing containment tests on these lists it might be better to convert them into [sets](https://docs.python.org/2/library/sets.html) for faster lookup time. It probably won't make a huge difference though unless someone goes crazy with the custom statuses.

RouganStriker:
Looks good :+1:

sdetools/alm_integration/alm_plugin_base.py
lmontopo: Good observation.  I'll go ahead and make that change because I don't see any harm in being proactive.  Thanks!

PR #340, opened 2016-02-29
Update sdetools created task notes to show status meaning

lmontopo:
* Added a check to see which type of meaning a task had, and printed the meaning instead of the task status id for more user friendly notes.

Refs: #SDE-2120


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -650,9 +650,12 @@ def sde_update_task_status(self, task, status):
                                'Elements. Either the task no longer '
                                'exists, there was a problem connecting '
                                'to the server, or the status was invalid')
+        readable_status = 'DONE'
+        if status in STATUS.TODO_SET:
+            readable_status = 'TODO'
```
RouganStriker: According to the function comments it is possible to set the status as NA so lets add another conditional for that?

lmontopo: OK yea i was clumping that into DONE but good call on making it separate.

RouganStriker:
+1

PR #341, opened 2016-03-04
Fix JIRA JQL parameters

RouganStriker:


geoffwhittington:
Looks good, but I haven't tested it with the ALM title option. If it works with that, merge away!

RouganStriker:
I tested with an alm title option that began with a [ and it worked

PR #343, opened 2016-03-09
Support arbitrary field assignment in Rally sync

geoffwhittington:
This PR does the following:

- Renames Rally to "CA Agile Central" as per recent acquisition by CA
- Enables arbitrary field assignment that is possible in Jira
- Some configuration validation clean-up

PR #344, opened 2016-03-10
Support Rally Priority and Rename to CA Agile Central

geoffwhittington:


geoffwhittington:
@emintham This replaces https://github.com/sdelements/sdetools/pull/343 For some reason I couldn't resolve conflicts in that PR - but its done here.

RouganStriker:
Is priority a newly implemented field in Rally?

geoffwhittington:
Priority was always in Rally - however it is only there by default for Defects. Since our integration is focused on Story and Tasks we never implemented this. We have a customer who wants to use Priority even on Story/Tasks - so this is why this PR implements the capability generally.

RouganStriker:
Maybe I missed it but are we checking if the priority field is available for the specified Rally task type?

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -408,22 +443,29 @@ def alm_get_task(self, task):
...
                 'Name': task['alm_full_title'],
-                'Description': self.sde_get_task_content(task),
                 'Workspace': self.workspace_ref,
                 'Project': self.project_ref
             }
         }
 
+        for field_name, field_value in new_artifact.iteritems():
+
+            if field_name in self.artifact_fields:
```
geoffwhittington: @RouganStriker I think this should be checking that the "Priority" field is in the Task

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -863,3 +863,60 @@ def translate_priority(self, priority):
             else:
                 if int(key) == priority:
                     return pmap[key]
+
+
+class AlmAdvConnector(AlmConnector):
```
emintham: What's AlmAdv supposed to mean? ALMAdvanced? This class also needs a docstring.

geoffwhittington: Thank-you for seeing this. I have no idea what to call this class. This is the class that allows integrations to use macros inside custom fields... Any ideas for a good name? :)

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -863,3 +863,60 @@ def translate_priority(self, priority):
...
+        issue_template = {}
+        issue_template.update(self.default_issue_template)
+
+        if self.config['alm_custom_fields']:
+            issue_template.update(self.config['alm_custom_fields'])
+
+        return issue_template
+
+    def get_new_issue(self, task):
+        new_issue = {}
```
emintham:
```python
return dict((field, self._transform_value(value, task))
            for field, value in self.new_issue_template.iteritems())
```

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -863,3 +863,60 @@ def translate_priority(self, priority):
...
+
+        # each macro is assumed to be a string or a list of strings
+        task['__macros'] = {
+            'business_unit': self.config['sde_businessunit'],
+            'application': self.config['sde_application'],
+            'project': self.config['sde_project'],
+            'context': self.config['alm_context'],
+            'alm_user': self.config['alm_user'],
+            'id': task['id'],
+            'task_id': task_id,
```
emintham: Maybe a comment to explain how this task_id is different from task['id'] above it?

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -84,10 +98,12 @@ def get_timestamp(self):
         return parse(self.timestamp)
 
 
-class RallyConnector(AlmConnector):
+class RallyConnector(AlmAdvConnector):
     """Connects SD Elements to Rally"""
```
emintham: Maybe add a comment to this docstring that Rally is now CA Agile Central.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -863,3 +863,60 @@ def translate_priority(self, priority):
             else:
                 if int(key) == priority:
                     return pmap[key]
+
+
+class AlmAdvConnector(AlmConnector):
+    default_issue_template = {}
+
+    def get_issue_template(self):
```
emintham: Maybe change this to
```python
@property
def new_issue_template(self):
    # everything as before...
```

Then in the other part of the code, you don't have to say `self.new_issue_template = self.get_issue_template()`

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -113,6 +129,8 @@ def __init__(self, config, alm_plugin):
     def initialize(self):
         super(RallyConnector, self).initialize()
 
+        self.new_issue_template = self.get_issue_template()
```
emintham: Here

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -299,20 +317,37 @@ def alm_validate_configurations(self):
             raise AlmException('Error while retrieving attribute definitions for "%s": %s' %
                                (card_type_details['type'], err))
 
+        #  Retrieve artifact fields
+        self.artifact_fields = {}
```
emintham:
```python
self.artifact_fields = dict((attribute['Name'], attribute)
                             for attribute in requirement_attrs['TypeDefinition']['Attributes'])
```

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -299,20 +317,37 @@ def alm_validate_configurations(self):
...
+                               (self.config['rally_new_status'], ', '.join(allowed_values)))
 
-                difference_set = set(self.config['rally_done_statuses']).difference(allowed_values)
-                if difference_set:
-                    raise AlmException('Invalid rally_done_statuses %s. Expected one of %s' %
-                                       (difference_set, allowed_values))
-                return
-        raise AlmException('Unable to retrieve allowed values for "%s"' % card_type_details['name'])
+        # Validate the done stasuses
+        for done_status in self.config['rally_done_statuses']:
```
emintham: I know we need to maintain backwards compatibility but maybe keeping "rally" in the name isn't the most user-friendly option. Perhaps introduce new configs but check for both in the code with the new config taking precedence.

geoffwhittington: This would require a change to the SDE frontend as well - providing some sort of config migration capability and then re-writing the HTML with the new keys. I don't have the bandwidth for this

emintham: File a ticket :)

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -299,20 +317,37 @@ def alm_validate_configurations(self):
...
+            if done_status not in allowed_values:
+                raise AlmException('Invalid rally_done_statuses value: %s. Expected one of %s' %
+                                   (done_status, ', '.join(allowed_values)))
+
+        # Validate user-defined custom fields are in the project
+        for key in self.config['alm_custom_fields']:
```
emintham: I'd probably do a set difference here and report on all undefined fields in one go.

sdetools/alm_integration/alm_plugin_base.py
geoffwhittington: Sounds better than mine, I'll steal it with pride

emintham:
Seems mostly good. Left some minor comments.

sdetools/modules/sync_rally/tests/rally_plugin_test.py
```diff
@@ -26,7 +26,7 @@ def test_parsing_alm_task(self):
     def test_subscription_error(self):
         self.mock_alm_response.set_response_flags({'get_subscription': '401'})
 
-        self.assert_exception(AlmException, '', 'Unable to retrieve subscription from Rally.',
+        self.assert_exception(AlmException, '', 'Unable to retrieve subscription.',
```
geoffwhittington: I'll leave that for another merge ;)

sdetools/modules/sync_rally/rally_plugin.py
RouganStriker: Ah I see, so if the priority field is not available it won't get added to the post data

geoffwhittington:
@emintham ok updated, I'm going to bump sdetools version and merge

PR #347, opened 2016-03-11
Fix/optimize ldap

emintham:


emintham:
Ignore the broken tests for now. Haven't updated them to reflect code changes yet.

emintham:
Locally, these are the performance improvements observed:

no opt: 0.29s user 0.19s system 7% cpu 6.052 total
opt   : 0.29s user 0.17s system 7% cpu 5.890 total
opt2  : 0.22s user 0.11s system 10% cpu 3.179 total
opt3  : 0.21s user 0.12s system 17% cpu 1.887 total

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -319,28 +325,22 @@ def get_groups_from_ldap(self):
...
-            member = member.split(',')[0]
-            query = '(' + escape_filter_chars(member) + ')'
+        logger.debug("Getting members of group: %s", group)
 
-        user = self._ldap_search(query, self.get_ldap_user_params())
+        query = "(&(objectClass=user)(memberOf=%s))" % group
+        results = self._ldap_search(query, self.get_ldap_user_params())
 
-        if len(user):
+        if len(results):
```
JeffBain: Is this just a list? Any reason we don't just go 

    if results:

?

Likewise, do we want to do

if not results:
    return []

to avoid having to indent the subsequent logic?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -642,16 +650,25 @@ def sync_intersection(self, intersection, sd_users, directory_users):
...
 
-            if status == DEACTIVATED:
-                deactivated_user_count += 1
-            elif status == UPDATED:
-                updated_user_count += 1
+                    if status == DEACTIVATED:
+                        deactivated_user_count += 1
+                    elif status == UPDATED:
+                        updated_user_count += 1
+                except Exception:
```
JeffBain: Do we have a more specific Exception we can catch? I'm pretty wary about "Catch EVERYTHING." If we're expecting a specific kind of error.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -690,22 +719,34 @@ def sync_new_users_from_ldap(self, untracked_users, directory_users):
...
+                user_groups = self.translate_ldap_group_to_sde_group(dir_user["groups"],
+                                                                     self.config['group_mapping'])
+                groups = [grp['id'] for grp in user_groups]
+
+                name = self.calculate_name(dir_user)
+                dir_user.update(**name)
+                logger.debug("Creating new SD Elements user '%s'", user)
+
+                try:
+                    self.sde_plugin.create_user(email=user,
```
JeffBain: I don't suppose there's a clever way to bulk_create users? If we could validate them inline and then use bulk_create behind the scenes it might reduce the amount of transit time for this kind of stuff. Probably a lot of work though, since I assume this is talking through APIs and things.

JeffBain:
A lot of the code is reasonable, and if I understand it the LDAP filtering stuff seems to make sense. It feels like we need more options on our side to make these sorts of mass-update type actions easier. I wonder how many crazy queries we're doing with a big run of this.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -390,11 +392,14 @@ def parse_groups(self, ldap_groups):
...
         groups = defaultdict(list)
-        for group in ldap_groups:
-            group_name, members = self.get_group_from_ldap_tuple(group)
+        ldap_groups = (self.get_group_from_ldap_tuple(group)
+                       for group in ldap_groups)
+        for group_name, members, group_dn in ldap_groups:
 
-            # filter falsy users (not found in LDAP)
-            members = filter(bool, [self.get_ldap_user(user) for user in members])
+            if not group_name or group_name not in mapping.keys():
```
funkaoshi: You can just say `group_name not in mapping` I believe. The `keys` is implicit.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Fixed.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Fixed.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Not as far as I know. Everything is through API.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Done.

emintham:
@JeffBain One other alternative is to make this code export a CSV and then they can use the bulk import feature with the generated CSV. It does take more clicks though.

PR #350, opened 2016-03-16
ThreadFix integration

geoffwhittington:


sdetools/modules/import_threadfix/threadfix_integrator.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class ThreadFixAPI(RESTBase):
+
+    def __init__(self, config):
+        extra_conf_opts = [('analysis_api_token', 'ThreadFix API Token', '')]
+        super(ThreadFixAPI, self).__init__('analysis', 'ThreadFix', config, extra_conf_opts=extra_conf_opts)
+
+    def call_api(self, target, method=URLRequest.GET, args={}, call_headers={}):
+        url = '%s?apiKey=%s' % (target, self.config['analysis_api_token'])
```
RouganStriker: There is potential for this to cause call_api to generate an incorrect URL because we url encode the args in for a GET request.

I'm all for either creating a new url-based authentication method or expanding on the custom auth option you added to RESTBase. The expansion option could be something like
```
elif auth_mode == 'custom':
    req = self.custom_authentication(req)
```
Then we define custom_authentication in RESTBase to be something like
```
def custom_authentication(self, req):
    raise NotImplementedError
```

sdetools/sdelib/restclient.py
```diff
@@ -187,7 +187,10 @@ def encode_post_args(self, args, call_headers={}):
               # we just pass it through
             encoded_args = args
 
-        return json.dumps(encoded_args)
+        if 'Content-Type' in call_headers and call_headers['Content-Type'] == 'application/x-www-form-urlencoded':
```
RouganStriker: I think we can just move this conditional to the top of this function since the return value does not even use encoded_args.

Also, I'm not sure why we include call_headers as a parameter to this function. I think passing the content type would be more appropriate.

geoffwhittington: hmm.. I should be using encoded_args? otherwise I'm going to get into trouble with unicode post args

geoffwhittington: Unicode is my Achilles heel

sdetools/modules/import_threadfix/threadfix_integrator.py
geoffwhittington: The problem with a `custom_authentication` method is that it is too late. Once the URLRequest is made in the `call_api` method we can't go back and change it. What we'd end up doing is adding a hook so that a subclass could update the url, args or headers just before the URLRequest is made. But this is precisely what `call_api` allows subclasses to do.

geoffwhittington: The issue with ThreadFix is we have to set the apiKey in the URL parameters - nowhere else. If we put them in the args and we do a POST we will end up putting them into the data payload, which won't work unfortunately.

geoffwhittington: Maybe the fix is to add support for URL args

geoffwhittington: @emintham what do you think we should do here?

sdetools/sdelib/restclient.py
RouganStriker: I think its fine here, self.urlencode() should handle the unicode stuff as stuff.

geoffwhittington: do you mean its fine to move the conditional to the top of the function?

RouganStriker: Yea, before all the ```isinstance``` checks on ```args```

geoffwhittington:
@emintham can you check this over and leave your comments please?

emintham:
@geoffwhittington I'm in Japan right now... This MR is pretty large.. I can't really comment on the two issues raised above without looking more thoroughly at the MR and its related code.... Can you get someone else to look it over or can this wait till I get back?

geoffwhittington:
@emintham i completely forgot you were away! Sorry bud, enjoy.

funkaoshi:
@geoffwhittington can @RouganStriker rap this up?

PR #351, opened 2016-03-21
Allow specifying group_member_query

emintham:
- Script users can now specify a query to get all members of a given
  group.
- Prompt user before deactivating current users in SDE.


funkaoshi:
Seems good. Didn't look at why Travis is failing.

emintham:
That's because I haven't updated the tests to reflect the optimizations.

PR #352, opened 2016-03-22
Fix AppScan Enterprise triage capability and replace project referenc…

geoffwhittington:
…e with application


geoffwhittington:
@funkaoshi I'm merging this in to complete the AppScan Enterprise work

funkaoshi:
Sounds good. Let's get it in for v3.7

sdetools/modules/import_appscan/appscan_enterprise_xml_importer.py
```diff
@@ -70,7 +70,10 @@ def __init__(self):
         self.edition = 'enterprise'
 
     def _get_content_handler(self):
-        return AppScanEnterpriseXMLContent()
+        content_handler = AppScanEnterpriseXMLContent()
+        content_handler.blacklist_statuses = self.blacklist_statuses
```
emintham: I'd suggest having `blacklist_statuses` be an optional parameter that you pass in to the constructor instead of mutating it halfway before returning.

```python
return AppScanExterpriseXMLContent(blacklist_statuses=self.blacklist_statuses

# ...
class AppScanEnterpriseXMLContent(...):
    blacklist_statuses = ...
    def __init__(self, blacklist_statuses=None):
        self.blacklist_statuses = blacklist_statuses or AppScanEnterpriseXMLContent.blacklist_statuses
```

geoffwhittington: Interesting idea. Blacklists are handled differently in Fortify so maybe we can do this sort of thing as part of a bigger refactor

PR #353, opened 2016-03-23
Allow users to supply their own group query

emintham:
- Users can now specify the query to get the group
  that contains all relevant groups instead of just
  getting all entities that are of objectClass=group.
- Minor refactoring of prompt before deactivation--
  fixes prompting with deactivated users.


JeffBain:
+1. Code looks pretty straightforward to me. Good stuff.

emintham:
Thanks @JeffBain

funkaoshi:
This looks good!

PR #357, opened 2016-04-19
Throw an error when a project in Checkmarx has not yet been scanned

geoffwhittington:


geoffwhittington:
@sibartlett I'm going to merge this in. We are getting empty responses when a scan hasn't been made. I'm assuming people would rather abort the integration than continue

PR #362, opened 2016-05-12
SDETools Task Reference support

shawngustaw:


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -792,7 +827,26 @@ def synchronize(self):
...
+                    if not reference_data:
+                        reference_data = self._create_reference_data(sde_task['id'],
+                                                                     alm_task,
+                                                                     connection)
+
+                    reference_id = reference_data.pop('id')
+
+                    # check if Task Reference is out of date
+                    new_reference_data = alm_task.get_reference_dict(connection)
+                    reference_has_changed = new_reference_data != reference_data
```
RouganStriker: Does this need to be a separate variable? I think we can just do `new_reference_data != reference_data`

sdetools/modules/sync_jira/jira_plugin.py
```diff
@@ -151,6 +151,15 @@ def alm_add_task(self, task):
 
         url = self._get_issue_url(new_issue['key'])
 
+        data = {
+            'context': connection.id,
```
RouganStriker: What should happen when connection is None? Should we set the context to None or should we not add a task reference?

notsrg: I definitely want to refactor this out so I don't have to pass around the connection all the time.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +725,95 @@ def _determine_latest(self, alm_task, sde_task):
...
+        return data
+
+    def _get_reference_data(self, sde_task, alm_task):
+        reference_data = None
+        references = self.sde_plugin.get_task_references(sde_task)
+
+        # no references, create one
+        if not references:
+            data = self._make_reference_dict(alm_task)
+            return self.sde_plugin.add_task_reference(sde_task, data)
```
funkaoshi: I'd assign this to reference_data and then always return at the end of the method. Or i'd drop the else, so it's clearer that the if ends the method early.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +725,95 @@ def _determine_latest(self, alm_task, sde_task):
...
+
+    def _remove_alm_task(self, sde_task):
+        """
+        Given a task in SDE, remove it from an external system.
+        """
+
+        alm_task = self.alm_get_task(sde_task)
+        if alm_task:
+            self.alm_remove_task(alm_task)
+            # TODO: remove TaskReference object
```
funkaoshi: If you remove the task, the reference should cascade delete.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +725,95 @@ def _determine_latest(self, alm_task, sde_task):
...
+        }
+
+    def _create_reference_data(self, task_id, task):  # TODO: rename this to create_task_refernce
+        data = self._make_reference_dict(task)
+        self.sde_plugin.add_task_reference(task_id, data)
+        return data
+
+    def _get_reference_data(self, sde_task, alm_task):
+        reference_data = None
+        references = self.sde_plugin.get_task_references(sde_task)
```
funkaoshi: We make an "internal" API call per task? We might want to have all of this information come back as part of the tasks data. Something like `/api/v2/project/123/tasks/?include=external_references` perhaps. Though, maybe this can be done as some sort of clean up of how we sync in general. My sense is the process does more API calls than we may need to?

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +725,95 @@ def _determine_latest(self, alm_task, sde_task):
...
+
+        # no references, create one
+        if not references:
+            data = self._make_reference_dict(alm_task)
+            return self.sde_plugin.add_task_reference(sde_task, data)
+        else:
+            # find the correct existing reference
+            for ref in references:
```
funkaoshi: A (perhaps) idiomatic way to write these sorts of expressions is:

```
reference_data  = next((ref for ref in references if ref['context'] == self.connection.id), None) 
```

Ask someone else what they think.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +725,95 @@ def _determine_latest(self, alm_task, sde_task):
...
+        Given a task in SDE, remove it from an external system.
+        """
+
+        alm_task = self.alm_get_task(sde_task)
+        if alm_task:
+            self.alm_remove_task(alm_task)
+            # TODO: remove TaskReference object
+
+    def _create_or_update_reference(self, sde_task, alm_task):
+        reference_data = self._get_reference_data(sde_task,
```
funkaoshi: I'd unwrap these lines. It looks awkward, and I can't imagine the extra 8 characters will kill us.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +725,95 @@ def _determine_latest(self, alm_task, sde_task):
...
+                                                  alm_task)
+
+        if not reference_data:
+            reference_data = self._create_reference_data(sde_task,
+                                                         alm_task)
+
+        reference_id = reference_data.pop('id')
+
+        # check if Task Reference is out of date
```
funkaoshi: It won't be out of date if we just created right? May want to check for that.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +725,95 @@ def _determine_latest(self, alm_task, sde_task):
...
+        """
+        What takes precedence in case of a conflict of status. Starts with ALM
+        """
+
+        precedence = 'alm'
+        updated_system = 'SD Elements'
+
+        if (alm_task.get_status() in STATUS.DONE_SET or
+                alm_task.get_status() in STATUS.NA_SET):
+
```
funkaoshi: Drop the blank line.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +725,95 @@ def _determine_latest(self, alm_task, sde_task):
...
+        return reference_id
+
+    def _apply_conflict_policy(self, task_id, sde_task, alm_task):
+        """
+        What takes precedence in case of a conflict of status. Starts with ALM
+        """
+
+        precedence = 'alm'
+        updated_system = 'SD Elements'
+
```
funkaoshi:
```
status = alm_task.get_status()
if status in STATUS.DONE_SET or status in STATUS.NA_SET:
   ...
```

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +725,95 @@ def _determine_latest(self, alm_task, sde_task):
...
+            precedence = self._determine_latest(alm_task, sde_task)
+
+        if precedence == 'alm':
+            self.sde_update_task_status(sde_task, alm_task.get_status())
+        else:
+            self.alm_update_task_status(alm_task, sde_task['status']['id'])
+            updated_system = self.alm_name
+            updated_status = sde_task['status']['meaning']
+
+        self.emit.info(('Updated status of task {0} in {1} to {2}'
```
funkaoshi: should be a unicode string.

sdetools/modules/sync_jira/__init__.py
```diff
@@ -22,16 +22,17 @@ def configure(self):
         alm_api = JIRARestAPI(self.config)
         self.jira = JIRAConnector(self.config, alm_api)
 
-    def handle(self):
+    def handle(self, connection=None):
```
funkaoshi: Why pass in the connection object if we just need/use the ID? If you are running this not as a library inside sde would this be something you'd be able to pass in?

sdetools/sdelib/mod_mgr.py
```diff
@@ -160,7 +160,8 @@ def load_modules(plugins_path=None):
 
 
 def run_command(cmd_name, args, call_src, call_options={},
-        call_back=stdout_callback, call_back_args={}, plugins_path=None):
+                call_back=stdout_callback, call_back_args={},
+                plugins_path=None, connection=None):
```
funkaoshi: See comment above.

sdetools/alm_integration/alm_plugin_base.py
geoffwhittington: I think this is fine, `alm_remove_task` removes the task from Jira - not from SDE

sdetools/alm_integration/alm_plugin_base.py
geoffwhittington: Yeah we'll want external_references everytime we sync - but not everytime the tasks are retrieved via API, an include filter would be amazing

sdetools/modules/sync_jira/__init__.py
geoffwhittington: sdetools uses the command pattern - I'm not sure passing this into the `handle` method is in line with that. This will need to come as an option, defined in the `alm_plugin_base` - something like ?

```
        self.config.opts.add('alm_reference_context', 'Context to use to track external task references',
                default='')
```

sdetools/alm_integration/alm_plugin_base.py
notsrg: @geoffwhittington @funkaoshi I've added an include filter for references to the tasks endpoint in my other MR. Will update this stuff in a sec to use it.

funkaoshi: That's great.

sdetools/modules/sync_jira/tests/jira_plugin_test.py
```diff
@@ -19,6 +19,7 @@
 class JiraBaseCase(AlmPluginTestBase):
     @classmethod
     def setUpClass(cls):
+        JIRAConnector.supports_references = False
```
shawngustaw: @geoffwhittington Is this terrible to do? By default it's set to True so when you run things normally it'll do all the reference checks, but for the tests they just get skipped.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -349,14 +355,16 @@ def alm_connect_project(self):
         pass
 
     @abstractmethod
-    def alm_get_task(self, task):
+    def alm_get_task(self, task, reference=None):
```
geoffwhittington: You don't need this. It is already available as `self.config['alm_reference_context']`

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -251,6 +255,8 @@ def initialize(self):
             if match not in AlmConnector.FIELD_OPTIONS:
                 raise AlmException('Incorrect alm_title_format configuration. Invalid field: ${%s}' % match)
 
+        self.alm_reference_context = self.config['alm_reference_context']
```
geoffwhittington: you can refer to the value by `self.config['alm_reference_context']`

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +733,103 @@ def _determine_latest(self, alm_task, sde_task):
...
+        alm_task = self.alm_get_task(sde_task)
+        if alm_task:
+            self.alm_remove_task(alm_task)
+        if task_reference and self.supports_references:
+            self.sde_plugin.remove_task_reference(sde_task, task_reference['id'])
+
+    def _reference_has_changes(self, old_reference, new_reference):
+        new_reference = self._make_reference_dict(new_reference)
+        old_reference.pop('id')
+        if old_reference != new_reference:
```
RouganStriker: I think we can simplify this to `return old_reference != new_reference`

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -493,6 +538,8 @@ def alm_add_task(self, task):
         if self.config['alm_standard_workflow'] and (task['status']['id'] in STATUS.DONE_SET or task['status']['id'] in STATUS.NA_SET):
             self.alm_update_task_status(alm_task, task['status']['id'])
 
+        return alm_task
```
RouganStriker: Is this a typo? Or do we not need the return statement below anymore?

sdetools/modules/sync_rally/tests/rally_plugin_test.py
```diff
@@ -10,6 +10,8 @@
 class TestRallyCase(AlmPluginTestBase, unittest.TestCase):
     @classmethod
     def setUpClass(cls):
+        RallyConnector.supports_references = False
```
geoffwhittington: I'd expect that Rally should support references

sdetools/modules/sync_rally/rally_plugin.py
shawngustaw: @RouganStriker return statement below should've been deleted.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +733,103 @@ def _determine_latest(self, alm_task, sde_task):
...
+        """
+        Given a task in SDE, remove it from an external system.
+        """
+        alm_task = self.alm_get_task(sde_task)
+        if alm_task:
+            self.alm_remove_task(alm_task)
+        if task_reference and self.supports_references:
+            self.sde_plugin.remove_task_reference(sde_task, task_reference['id'])
+
+    def _reference_has_changes(self, old_reference, new_reference):
```
geoffwhittington: What is the definition of a reference change? Is it one or all of the task references?

Is this method more accurately called `_references_have_changed`?

shawngustaw: @geoffwhittington still refining this, but I think alm_connection, task and external_id should be set in stone, and has_changes would be whether url_reference or user_friendly has changed. Basically this is the case where they move the task into a different project.

shawngustaw: (The code doesn't currently reflect that, you can change almost anything on a reference at the moment I believe)

sdetools/alm_integration/alm_plugin_base.py
shawngustaw: @geoffwhittington this is the actual Task Reference, not the ALM context, so it likely will have to get passed in.

geoffwhittington: The connector knows the context and `task` has the references in it, I believe. Can't the reference be derived from these two?

sdetools/modules/sync_jira/jira_shared.py
```diff
@@ -36,3 +41,11 @@ def get_status(self):
     def get_timestamp(self):
         """ Returns a datetime object """
         return parse(self.timestamp)
+
+    def get_reference_dict(self, connection):
```
shawngustaw: This might not be used anymore.

sdetools/alm_integration/alm_plugin_base.py
shawngustaw: agh you're right.

geoffwhittington: I see what your doing, and its good. But we can probably make a parent method that gets the right reference without affecting this method signature, which helps reduce the impact on the other tools

sdetools/modules/sync_trac/trac_plugin.py
```diff
@@ -186,6 +186,9 @@ def alm_add_task(self, task):
 
         if self.config['alm_standard_workflow'] and (task['status']['id'] in STATUS.DONE_SET or task['status']['id'] in STATUS.NA_SET):
             self.alm_update_task_status(alm_task, task['status']['id'])
+        return alm_id
```
RouganStriker: return alm_tasl

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -344,8 +353,14 @@ def alm_validate_configurations(self):
...
 
-    def rally_get_artifact(self, name, query, card_type, artifact_type, api):
+    def rally_get_artifact_by_reference(self, reference):
+        try:
+            result = self.alm_plugin.call_api('hierarchicalrequirement/%s.js' % reference)
+        except APIError:
+            raise Exception('Unable to lookup Rally task via its reference')
+        return result
 
+    def rally_get_artifact(self, name, query, card_type, artifact_type, api):
```
RouganStriker: This isn't returning anything

sdetools/modules/sync_github/github_plugin.py
```diff
@@ -276,6 +276,9 @@ def alm_add_task(self, task):
         if (self.config['alm_standard_workflow'] and task['status']['id'] in STATUS.DONE_SET or task['status']['id'] in STATUS.NA_SET):
             self.alm_update_task_status(alm_task, task['status']['id'])
 
+        return alm_task
+
+    def get_logging_message(self, alm_task):
```
geoffwhittington: Technically, we use these strings in the task notes (when a task is added to an ALM) and not in any logging. I'm not sure if this will be confused to future developers or not if you call it a logging message.

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -408,7 +423,37 @@ def _apply_labels_if_needed(self, labels, artifact_data):
...
+
+        return RallyTask(task_id,
+                         task_data['FormattedID'],
+                         self._split_ref_link(task_data['_ref']),
+                         task_data[card_type_details['field_state']],
+                         task_data['LastUpdateDate'],
+                         self.config['rally_done_statuses'],
+                         self._split_ref_link(task_data['_ref']),   # url ref
+                         task_data['ObjectID'],  # external id
+                         task_data['FormattedID'],  # user friendly
```
shawngustaw: NTS: A bunch of this gets passed in twice. Probably should fix that.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -754,91 +876,81 @@ def synchronize(self):
...
-                        self.alm_remove_task(alm_task)
-                    progress += 1
-                    self.output_progress(100*progress/total_work)
+                total_work = progress.count + len(tasks)
 
             for sde_task in tasks:
-                tid = self._extract_task_id(sde_task['id'])
-                progress += 1
-                self.output_progress(100*progress/total_work)
+                task_id = self._extract_task_id(sde_task['id'])
```
geoffwhittington: We don't need `_extract_task id()` anymore since the API returns the value now: `sde_task['task_id']`. I understand this is existing code.

sdetools/modules/sync_jira/jira_plugin.py
```diff
@@ -27,6 +28,8 @@ class JIRAConnector(MacroAlmConnector):
     default_issue_template = JIRA_DEFAULT_ISSUE_TEMPLATE
     default_priority_map = JIRA_DEFAULT_PRIORITY_MAP
 
+    supports_references = True
```
geoffwhittington: Does Jira work when you set the version == 4 ? This will test SOAP

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -155,7 +155,61 @@ def get_issue_types(self):
...
+    def get_task(self, task, task_id, alm_task_id=None):
+        # Get by reference if provided
+        if alm_task_id:
+            args = {
+                'fields': 'resolution,versions,priority,status,updated,created'
+            }
+            try:
+                result = self.call_api('issue/%s' % alm_task_id, args=args)
+            except APIError, error:  # likely a stale Task Reference
+                return None
```
geoffwhittington: We can't assume an API error is a stale task reference. We may get an error given any number of reasons; including server is down, bad credentials, network is lost, etc.

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -155,7 +155,61 @@ def get_issue_types(self):
...
+            return self.create_task_obj(task_id, result)
 
         task_lookup = []
```
geoffwhittington: Add a comment about what condition we're in:

```
# No reference is associated with the task, look for task using legacy approach: title and custom lookup fields
```

sdetools/modules/sync_jira/tests/jira_plugin_test.py
geoffwhittington: Assuming I understand things correctly, I don't think you want to do this, doesn't it avoid the code you are introducing in this PR? Are you trying to avoid updating the tests?

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -344,8 +353,14 @@ def alm_validate_configurations(self):
         if unexpected_custom_fields:
             raise AlmException('Specified fields not found in the project: %s' % ', '.join(unexpected_custom_fields))
 
-    def rally_get_artifact(self, name, query, card_type, artifact_type, api):
+    def rally_get_artifact_by_reference(self, reference):
+        try:
+            result = self.alm_plugin.call_api('hierarchicalrequirement/%s.js' % reference)
```
geoffwhittington: What does a Rally reference look like? I expected it would look like a URL

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -408,7 +423,37 @@ def _apply_labels_if_needed(self, labels, artifact_data):
             raise AlmException('Unable to update tag %s for artifact %s because of %s' %
                                (tag_name, artifact_data['FormattedID'], err))
 
+    def create_obj_from_data(self, task, task_data):
```
geoffwhittington: It would be good to rename this to something like `create_alm_task_from_data` perhaps? - "obj" seems a little vauge

sdetools/modules/sync_rally/rally_plugin.py
RouganStriker: It seems RallyTask already has a url reference field. But I think it would be okay to leave it as-is for now if refactoring this to remove duplicates is non-trivial. @geoffwhittington some of the fields in these alm task objects are unnecessary(e.g. in the JiraTask object it stores the priority and resolution value but this never gets used), it might be worthwhile to make a tech debt ticket to clean them up.

geoffwhittington: I'm always in favour of removing cruft

sdetools/modules/sync_jira/jira_shared.py
geoffwhittington: Delete and find out ;)

sdetools/modules/sync_jira/jira_plugin.py
geoffwhittington: As a follow-up, we don't need to support references with Jira 4 (SOAP)

sdetools/modules/sync_jira/jira_plugin.py
```diff
@@ -149,10 +160,13 @@ def alm_add_task(self, task):
...
             self.alm_update_task_status(alm_task, task['status']['id'])
 
-        url = self._get_issue_url(new_issue['key'])
-
-        # Return a unique identifier to this task in JIRA
-        return 'Issue %s, URL: %s' % (new_issue['key'], url)
+        return JIRATask(
+            task,
+            new_issue['key'],
+            link=new_issue['self'],
```
RouganStriker: This is a link to the json representation(eg https://sdetest.atlassian.net/rest/api/2/issue/46919). On the task notes we link directly to the issue(eg https://sdetest.atlassian.net/browse/KSYN3-10) so there is an inconsistency in the way we display the links.

geoffwhittington: The `link` should be a user-clickable web reference: i.e.  https://sdetest.atlassian.net/browse/KSYN3-10

RouganStriker: We can reuse ```url = self._get_issue_url(key)```

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -408,7 +426,37 @@ def _apply_labels_if_needed(self, labels, artifact_data):
...
+
+        card_type_details = self.card_types[self.config['rally_card_type']]
+
+        return RallyTask(task_id,
+                         task_data['FormattedID'],
+                         self._split_ref_link(task_data['_ref']),
+                         task_data[card_type_details['field_state']],
+                         task_data['LastUpdateDate'],
+                         self.config['rally_done_statuses'],
+                         self._split_ref_link(task_data['_ref']),   # url ref
```
RouganStriker: self._get_issue_url(alm_task) to get user friendly URL

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -183,6 +183,10 @@ def carriage_return(self):
         return '<br//>'
 
     def _get_issue_url(self, alm_task):
+        if type(alm_task) == RallyTask:
```
RouganStriker: Would it be simpler if we just replace the alm_task parameter with alm_task_ref? It doesn't seem like this function uses anything in the alm_task other than the alm_task_ref

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -725,6 +729,115 @@ def _determine_latest(self, alm_task, sde_task):
...
+        if alm_task:
+            self.alm_remove_task(alm_task)
+        if task_reference and self.supports_references:
+            self.sde_plugin.remove_task_reference(sde_task, task_reference['id'])
+
+    def _test_reference_out_of_date(self, old_reference, new_reference):
+        """
+        Test whether an old Task Reference is stale, i.e. Jira task
+        has moved projects, so its url reference is out of date.
+        Should pass in a copy to old ref to prevent side effects.
```
jnlantos: Might be good to make the copy of old ref here, so the caller doesn't have to know that they need to pass in a copy.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -754,91 +876,81 @@ def synchronize(self):
...
-                            precedence = self._determine_latest(alm_task, sde_task)
-
-                        if precedence == 'alm':
-                            self.sde_update_task_status(sde_task, alm_task.get_status())
-                        else:
-                            self.alm_update_task_status(alm_task, sde_task['status']['id'])
-                            updated_system = self.alm_name
-                            updated_status = sde_task['status']['meaning']
-                        self.emit.info('Updated status of task %s in %s to %s' % (tid, updated_system, updated_status))
+                        self._apply_conflict_policy(task_id, sde_task, alm_task)
```
jnlantos: Nice.

sdetools/sdelib/interactive_plugin.py
```diff
@@ -208,6 +208,22 @@ def add_task_text_note(self, task_id, text):
         return self.api.add_task_note(self.prj_id, task_id, text)
 
     @_verify_connect
+    def add_task_reference(self, task_id, data):
+        return self.api.add_task_reference(self.prj_id, task_id, **data)
+
+    @_verify_connect
+    def update_task_reference(self, task_id, reference_id, *args, **kwargs):
```
jnlantos: Where is args used here?

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -408,7 +430,37 @@ def _apply_labels_if_needed(self, labels, artifact_data):
...
+        # perform lookup by reference if one exists
+        if reference:
+            reference = reference.get('reference')
+            task_data = self.rally_get_artifact_by_reference(reference)
+            task_data = task_data.get('HierarchicalRequirement', None)
+
+            if not task_data:
+                return None
+            return self.create_task_obj(task, task_data)
+
```
jnlantos: If the stuff below is the legacy way, I'd put a similar comment here to the one you put in the jira plugin. It made things clearer there, so might be good to have here as well.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -754,91 +876,81 @@ def synchronize(self):
...
+                total_work = progress.count + len(tasks)
 
             for sde_task in tasks:
-                tid = self._extract_task_id(sde_task['id'])
-                progress += 1
-                self.output_progress(100*progress/total_work)
+                task_id = sde_task['task_id']
+                update_progress(1, 100 * progress.count / total_work)
+
+                task_reference = self._get_matching_reference(sde_task.copy())
```
jnlantos: Yeah, try just passing in sde_task here and using get like you mentioned.

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -155,8 +156,67 @@ def get_issue_types(self):
...
+            except APICallError as error:
+                return None  # likely a stale Task Reference
+            except APIError, error:
+                pass
+            else:
+                return self.create_task_obj(task_id, result)
 
+        # No reference is associated with the task, look for task using legacy
+        # approach: title and custom lookup fields
         task_lookup = []
```
jnlantos: I like how you refactored self.create_task_obj() to it's own method. What do you think about refactoring out the legacy stuff too (here and in rally), and making like a look_for_task_using_legacy() method (with a better name maybe)?

sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -174,15 +183,15 @@ def carriage_return(self):
         return '<br//>'
 
     def _get_issue_url(self, alm_task):
```
RouganStriker: Should rename this to alm_task_reference

sdetools/sdelib/interactive_plugin.py
shawngustaw: @jnlantos removed 💃

sdetools/modules/sync_jira/jira_plugin.py
```diff
@@ -149,10 +160,13 @@ def alm_add_task(self, task):
...
-        url = self._get_issue_url(new_issue['key'])
-
-        # Return a unique identifier to this task in JIRA
-        return 'Issue %s, URL: %s' % (new_issue['key'], url)
+        return JIRATask(
+            task,
+            new_issue['key'],
+            reference=new_issue['id'],
+            name=new_issue['key'],
+            link=new_issue['self']
```
RouganStriker: self._get_issue_url(new_issue['key']) to link it to the web page version of the task

funkaoshi: @RouganStriker you want to just commit the change? @jnlantos ^^^

PR #363, opened 2016-05-16
Fixed SNI for Python 2.7.9+

ehsanf:


emintham:
Why are we duplicating yantra code in sdetools? Is there a problem with using it as a dependency?

geoffwhittington:
@emintham sdetools stopped being portable when yantra became an external dependency (check the extlib directory, we have put a bunch of dependencies in there).

funkaoshi:
I also think we should get away from "vendoring" all our dependencies. Is the issue that yantra isn't up on pypi? That's one problem right now. You can't just pip install sdetools.

sdetools/sdelib/restclient.py
```diff
@@ -299,12 +299,22 @@ def call_api(self, target, method=URLRequest.GET, args=None, call_headers={}, au
...
+                    type(err.reason) is http_req.ssl.SSLError):
+                raise ServerError('Unable to verify SSL certificate for host: %s (reason: %s)' % 
+                        (self.server, err.reason))
             handle = err
             call_success = False
         except httplib.InvalidURL, err:
             raise UsageError('Invalid URL. Reason %s' % (err))
+        except Exception as err:
+            if http_req.ssl_lib_found and type(err) is http_req.ssl.CertificateError:
+                raise ServerError('Unable to verify SSL certificate for host: %s (reason: %s)' %
```
funkaoshi: Can make a new exception SSLError or something that takes self.server & err.reason as params so you don't have to repeat your message all over the place.

funkaoshi:
@emintham you may need to wrap this up. I think @ehsanf pushing this up at all was a one time thing. Ha

geoffwhittington:
The original practice of putting all its dependencies into the extlib directory allowed us to: 1) make sure it had everything it needed to run and 2) fork & deploy customized 3rd party libraries (SOAPpy for one).

PR #366, opened 2016-06-08
Workaround VSO collection bug

geoffwhittington:


geoffwhittington:
@funkaoshi this should do the trick. I tested it with my instance on VSO

funkaoshi:
+1 Ha.

PR #368, opened 2016-06-09
Fix versionone integration error caused by unencoded xml payload

RouganStriker:
Refs #SDE-2683


geoffwhittington:
@emintham does this look reasonable?

emintham:
+1

PR #374, opened 2016-07-20
Rally - Fixed custom lookup macros not being transformed

sibartlett:


sdetools/modules/sync_rally/rally_plugin.py
```diff
@@ -419,7 +419,8 @@ def alm_get_task(self, task):
                              artifact_query, self.config['alm_parent_issue'])
 
         for field, value in self.config['alm_custom_lookup_fields'].items():
-            artifact_query = '(%s and (%s = "%s"))' % (artifact_query, field, value)
+            transformed_value = self._transform_value(value, task['__macros'])
+            artifact_query = '(%s and (%s = "%s"))' % (artifact_query, field, transformed_value)
```
geoffwhittington: this looks fine - but does this work with `field` value having a space in it? We saw this with Vey. I guess this would be a different bug anyhow than the one you're fixing here.

geoffwhittington:
+1

PR #375, opened 2016-07-21
Tasks in SD with "accepted=False" are marked as done in ALM during a sync

keyvanm:
Tasks in SD with "accepted=False" are marked as done in ALM during a sync


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -796,6 +798,11 @@ def synchronize(self):
                     if not self.config['alm_standard_workflow']:
                         continue
 
+                    if not sde_task['accepted']:
```
funkaoshi: I'd add little comments about what's up. Eg. `# Close tasks that are no longer part of the SDE project.`

sdetools/sdelib/sdeapi.py
```diff
@@ -134,7 +134,7 @@ def get_projects(self, application, options={}, **filters):
                 call_headers=_encode_options(options))
         return response['results']
 
-    def get_tasks(self, project_id, args=None, **filters):
+    def get_tasks(self, project_id, args={}, **filters):
```
funkaoshi: Why the change here? Normally you shouldn't make python default arguments mutable objects.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -764,13 +764,15 @@ def synchronize(self):
             progress += 2
             self.output_progress(progress)
 
-            # Attempt to get all tasks
+            # Attempt to get all (Accepted) tasks
             tasks = self.sde_get_tasks()
```
sibartlett: Rename this variable to `accepted_tasks`.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -764,13 +764,15 @@ def synchronize(self):
...
+            unaccepted_tasks = self.sde_get_tasks(accepted=False)
             logger.info('Retrieved all tasks from SDE')
 
             # Filter tasks - progress must match reality
             tasks = self.filter_tasks(tasks)
```
sibartlett: Then change these 2 lines to:

```
tasks = self.filter_tasks(accepted_tasks + unaccepted_tasks)
```

sdetools/sdelib/sdeapi.py
keyvanm: If for some reason the arg is set to None, the first line of the method will throw an error.
```python
args.update({'include': 'tags,how_tos,problem,related', 'expand': 'status,phase', 'absolute_urls': True})
```

JeffBain: Seems like we need a

    if args is none:
        args = {}

sibartlett:
Merged in #405.

PR #377, opened 2016-07-25
Add project custom attributes macros

MartinPetkov:
Refs #SDE-2925


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -557,6 +557,23 @@ def get_alm_task_title(config, task, fixed=False):
...
+        if not self.sde_plugin:
+            raise AlmException('Requires initialization')
+
+        try:
+            return self.sde_plugin.get_project()
+        except APIError, err:
+            logger.error(err)
+            raise AlmException('Unable to get project from SD Elements. Please ensure'
+                               ' the project is valid and that the user has'
+                               ' sufficient permission to access the project. Reason: %s' % (str(err)))
```
JeffBain: Is there a reason we're still using the old style % formatting method instead of .format() in SDETools?

sdetools/sdelib/testlib/sde_response_generator.py
```diff
@@ -220,6 +221,13 @@ def get_projects(self, target, flag, data, method):
         projects = self.generator_get_filtered_resource('project', self.get_url_parameters(target))
         return RESPONSE_HEADERS, {'results': projects}
 
+    def get_project(self, target, flag, data, method):
```
JeffBain: It's sort of unclear what these parameters are for, maybe a docstring would be helpful?

Also do we use any of these in this method? It seems like we don't, so I'm wondering if it's just something non-obvious in the diff?

JeffBain: Could probably use a docstring to explain parameter stuff at least? Kind of unclear what's going on here.

sibartlett: All the mock requests use the same method signature.

Again, we can revisit this later as part of our refactoring efforts.

PR #380, opened 2016-07-28
RTC - Added support for specifying parent work item

sibartlett:


sdetools/modules/sync_rational/rational_plugin.py
```diff
@@ -293,6 +293,15 @@ def _get_rtc_field(self, field_name, field_value):
         if not field_name:
             raise AlmException('Unsupported attribute: %s' % attribute)
 
+        if rs['oslc:name'] == 'com.ibm.team.workitem.linktype.parentworkitem.parent':
```
geoffwhittington: This looks good. Better to not match against "Parent"

sdetools/modules/sync_rational/rational_plugin.py
```diff
@@ -293,6 +293,15 @@ def _get_rtc_field(self, field_name, field_value):
...
+            resource_url = self.alm_plugin.get_url('resource/itemName/com.ibm.team.workitem.WorkItem/%s' % field_value)
```
geoffwhittington: What is an example of "field_value"?

geoffwhittington: I am only concerned that `field_value` is an internal value and not a user-friendly one. The use case is developers will be entering the name of the Epic and I'm expecting that value should be obvious from the RTC UI

sibartlett: It's the numerical workitem ID, which is displayed next to the name of the Epic in RTC UI.

PR #381, opened 2016-08-04
Add flag for saving FPR to disk

shawngustaw:


sdetools/modules/import_fortify/fortify_ssc_importer.py
```diff
@@ -163,21 +164,25 @@ def run(self):
...
         logger.info('FPR file downloaded successfully')
 
         self.importer = FortifyFPRImporter(self.config['import_blacklist'])
-        self.importer.parse(temp_fd)
+        self.importer.parse(fd)
 
         # Cleanup temporary file
-        temp_fd.close()
+        if self.use_temp_file:
+            fd.close()
```
geoffwhittington: I would expect we want to close the file in either case - in the first case the temp file gets auto-deleted, in the latter case we get a saner file.

emintham:
+1

shawngustaw:
Moved this to a new PR with the correct base branch (version 5.2)

geoffwhittington:
@shawngustaw we don't actually want to merge this in I don't think. @sibartlett  ?

PR #383, opened 2016-08-10
Added custom fields and macro support to tfs_plugin

keyvanm:


sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -75,9 +75,12 @@ def get_timestamp(self):
         return parse(self.timestamp)
 
 
-class TFSConnector(AlmConnector):
+class TFSConnector(MacroAlmConnector):
     alm_name = 'TFS'
     default_priority_map = TFS_DEFAULT_PRIORITY_MAP
+    feature_custom_lookup = True
```
sibartlett: We don't need this line, it's for indicating we support `custom_lookup_fields`.

sibartlett:
The tests are failing. You can take a look at the Travis CI build to see why - I think the new API call needs to be implemented in the tests.

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -183,14 +194,13 @@ def _get_workitem_from_tfs(self, work_item_id, task_id):
...
     def alm_add_task(self, task):
-
         if self.config['alm_issue_label']:
             tags_string = self.config['alm_issue_label']
 
         for tag in task['tags']:
             tags_string += '; ' + tag
 
-        data = [
+        create_args = [
```
geoffwhittington: These default fields should be defined as a constant at the top of the file, similar to: https://github.com/sdelements/sdetools/blob/master/sdetools/modules/sync_rational/rational_plugin.py#L26

So we would get something like,

`
TFS_DEFAULT_ISSUE_TEMPLATE = {
   .....
}
`

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -183,14 +194,13 @@ def _get_workitem_from_tfs(self, work_item_id, task_id):
         return TFSTask(task_id, response['value'][0], self.config['tfs_done_statuses'])
 
     def alm_add_task(self, task):
-
```
geoffwhittington: The method should start out similar to https://github.com/sdelements/sdetools/blob/master/sdetools/modules/sync_rational/rational_plugin.py#L423 calling `new_artifact = self.get_new_issue(task['__macros'])`

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -75,9 +75,11 @@ def get_timestamp(self):
         return parse(self.timestamp)
 
 
-class TFSConnector(AlmConnector):
+class TFSConnector(MacroAlmConnector):
     alm_name = 'TFS'
     default_priority_map = TFS_DEFAULT_PRIORITY_MAP
```
geoffwhittington: Missing line of code similar to: https://github.com/sdelements/sdetools/blob/master/sdetools/modules/sync_rational/rational_plugin.py#L156

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -141,8 +143,17 @@ def alm_connect_project(self):
         if not found_project:
             raise AlmException('Project %s could not be found' % self.config['alm_project'])
 
+        self.tfs_work_item_fields_api_url = '{0}/_apis/wit/fields'.format(self.config['tfs_collection'])
```
geoffwhittington: Do we need to create this variable if we can just generate the URL when its needed? I guess this is a style thing...  I'll defer to @sibartlett

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -248,6 +260,13 @@ def alm_add_task(self, task):
 
         return work_item
 
+    def inject_custom_fields_into_create_args(self, task, create_args):
+        new_artifact = self.get_new_issue(task['__macros'])
```
geoffwhittington: Move this up to `alm_add_task`.

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -248,6 +260,13 @@ def alm_add_task(self, task):
 
         return work_item
 
+    def inject_custom_fields_into_create_args(self, task, create_args):
```
geoffwhittington: I'm not convinced you will need this function, you can probably move most/all of this into the `alm_add_task` method.

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -283,3 +302,38 @@ def alm_disconnect(self):
...
+        create_args_path = self.get_tfs_path_from_field_name(field_name)
+        create_args_value = field_value
+        return {
+                'op': 'add',
+                'path': create_args_path,
+                'value': create_args_value
+        }
+
+    @staticmethod
+    def modify_or_append(create_args, tfs_custom_field_arg):
```
geoffwhittington: Given the way I think `new_issue_template` works - there should never be a duplicate. I'm not sure we need this method.

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -283,3 +302,38 @@ def alm_disconnect(self):
 
     def convert_markdown_to_alm(self, content, ref):
         return self.mark_down_converter.convert(content)
+
+    def convert_sde_field_to_tfs_field(self, field_name, field_value):
```
geoffwhittington: Why not simplify the implementation to a single `return` statement?

geoffwhittington:
Looking good @keyvanm Looks like you've got it working 👍

sdetools/modules/sync_tfs/tfs_plugin.py
sibartlett: Yes, I agree. The URL should be generated when it is needed.

sdetools/modules/sync_tfs/tfs_plugin.py
keyvanm: It doesn't seem like rational itself uses this variable in a meaningful way. It uses `create_args` in its `alm_add_task` method: https://github.com/sdelements/sdetools/blob/master/sdetools/modules/sync_rational/rational_plugin.py#L425

sdetools/modules/sync_tfs/tfs_plugin.py
keyvanm: Just a personal preference for more readable code

sdetools/modules/sync_tfs/tfs_plugin.py
geoffwhittington: Yeah, consistency is a game we're not 100% good at yet. The goal is all fields in a sync are exclusively defined either by the DEFAULT_ISSUE_TEMPLATE variable or the custom fields specified in the config.

All the classes that inherit from `MacroAlmConnector` strive (loosely) to define the default set of fields that would be populated if there were no custom fields in a configuration. These default fields within a DEFAULT_ISSUE_TEMPLATE  variable are captured and communicated by the `new_issue_template` method (via self.default_issue_template) here: https://github.com/sdelements/sdetools/blob/master/sdetools/alm_integration/alm_plugin_base.py#L1010

Good eye, with your continued help we'll achieve a better consistency

sdetools/modules/sync_tfs/tfs_plugin.py
keyvanm: If they change one of the existing fields like `Title` in the custom mapping, we need to update the `create_args` with the new value. If you are certain that the `if len(fields) > 1:` case never happens, I can remove it.

geoffwhittington: If `create_args` is a dictionary you can always set fields as keys without worry, it will overwrite existing or add new. Then when `create_args` is done you can iterate over it all and build the list of values you ultimately want  to send to TFS.

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -283,3 +302,38 @@ def alm_disconnect(self):
...
+
+    @staticmethod
+    def modify_or_append(create_args, tfs_custom_field_arg):
+        fields = [field for field in create_args if field['path'] == tfs_custom_field_arg['path']]
+
+        if len(fields) == 0:
+            create_args.append(tfs_custom_field_arg)
+            return
+        if len(fields) == 1:
+            fields[0]['value'] = tfs_custom_field_arg['value']
```
geoffwhittington: Python blows my mind: is it pass by reference, pass by value? It amazes me that this works - but I don't doubt that it does.

@sibartlett Whats your preference here?

(background: Python is not my strong suit)

geoffwhittington:
@sibartlett I have a few outstanding comments, do you have anything to add?

sibartlett:
My suggestions:

1. Structure `create_args` like so:

```python
create_args = {
  'System.State': self.config['tfs_new_status'],
  'System.IterationPath': self.config['tfs_iterationpath']
}
```

2. Then the logic in `modify_or_append` is not needed, you just need to overwrite the keys in `create_args`.

3. Transform `create_args` to the format that TFS expects before passing to `self.alm_plugin.call_api`:

```python
[
    {
        'op': 'add',
        'path': '/fields/{0}'.format(field_name),
        'value': value
    }
    for field_name, value in create_args.items()
]
```

sibartlett:
Closed by #395.

PR #384, opened 2016-08-10
NTLM authentication support for TFS

geoffwhittington:
This seems to work. Needs to be updated so that we can support basic auth and ntlm - we might want to autodetect the situation somehow?


sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -22,6 +22,7 @@
 class TFSAPIBase(RESTBase):
     def __init__(self, config):
         super(TFSAPIBase, self).__init__('alm', 'TFS', config, '')
+        self.set_auth_mode('ntlm')
```
geoffwhittington: One thing we can do is in the TFSConnector.alm_connect_server() we can try to connect using basic auth and when it fails we can examine the response. If `WWW-Authenticate:NTLM` is in the header we can reset the auth mode to `ntlm`

geoffwhittington:
@sibartlett I turned off Basic Auth on the labs server and this seemed to work. Do you want to sanity check this too?

geoffwhittington:
See the build error? So funny, flake is complaining about flake.

sibartlett:
I've rebased this pull request against `version/5.2`.

I've updated the ntlm option to use basic auth with a "fallback" to NTLM. Seems to work for both our internal TFS and VSO.

geoffwhittington:
+1 if it works

PR #386, opened 2016-08-15
Convert PivotalTracker to MacroAlmConnector 

MartinPetkov:
Refs #SDE-3034


sdetools/modules/sync_pt/pt_plugin.py
```diff
@@ -338,6 +340,12 @@ def alm_add_task(self, task):
         if pt_labels:
             create_args['labels'] = pt_labels
 
+        if self.sync_titles_only:
+            create_args['description'] = PUBLIC_TASK_CONTENT
+
+        for field_name, field_value in new_artifact.iteritems():
+            create_args[field_name] = str(field_value)
```
sibartlett: Is there a particular reason we are forcing values to be string? Can't we accept the value as is?

```
create_args.update(new_artifact)
```

MartinPetkov: No there isn't, it's just leftover from the code I had before. I'll change it.

sdetools/modules/sync_pt/pt_plugin.py
```diff
@@ -338,6 +340,11 @@ def alm_add_task(self, task):
         if pt_labels:
             create_args['labels'] = pt_labels
 
+        if self.sync_titles_only:
+            create_args['description'] = PUBLIC_TASK_CONTENT
```
sibartlett: Sorry, I just realized that these 2 lines should be below `create_args.update(new_artifact)`.

sibartlett:
This looks good to me.

I moved the code concerning `PUBLIC_TASK_CONTENT` down to `MacroAlmConnector`.

Do you want to take a quick look at this @geoffwhittington?

geoffwhittington:
This generally looks good. One thing I'm not sure of is whether the 'description' key in the PT_DEFAULT_ISSUE_TYPE is the label on the user interface or the internal key used for filling the 'Description' value. I'm not really attached to this for Pivotal Tracker as it is not used much, but it's more important for other integrations.

sibartlett:
@geoffwhittington I've been thinking that for all of our integrations we should:

1. match on internal key by default
2. fallback to matching on user interface label, if no match was found in step 1

geoffwhittington:
The problem with that approach is it takes the power away from the user.
People spittle be able to look at their UI and know how to configure the
custom fields. Am I understanding you?

sibartlett:
@geoffwhittington the user would still be able to use UI strings, just it'd be a fallback. Alternatively, the internal keys could be the fallback.

A potential problem with UI labels - is i18n. What if different users are using different locales in their ALM (say French or Spanish) - and sdetools can't match using their label? The ability to match on internal key would be good in this situation.

I would prefer to try matching on internal key first, and then UI label - at least in systems like TFS and RTC where the internal keys have no chance of clashing with UI labels.

sibartlett:
Also, at least for the time-being, I think we should avoid hard-coding third-party UI labels/strings in sdetools.

Pivotal Tracker doesn't expose UI labels via an API - hence why we didn't include any kind of mapping in this pull request. I think the same probably applies to GitHub too.

geoffwhittington:
I don't think this should come up very often. The user crafting the JSON
custom fields should be aware of the connection user context. In the future
we're going to allow people to completely override the connector at the
project level so that should mitigate things.

The issue I saw again and again supporting integrations was that people
didn't know what the internal field names were, I was constantly creating
weird REST calls on long awkward calls so we could figure out what to put
in the JSON dictionary. I don't know a better way for this UX but I'm happy
you're thinking about it. Whatever you think is best.

PR #387, opened 2016-08-15
Support SD Elements versions prior to 4.0

sibartlett:


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -749,6 +749,16 @@ def _determine_latest(self, alm_task, sde_task):
 
         return 'sde' if sde_time > alm_time else 'alm'
 
+    def _check_supports_references(self, tasks):
+        if not tasks:
+            return
+
+        for task in tasks:
+            if 'references' in tasks:
```
geoffwhittington: `if 'references' in task:` I think?

sibartlett: Hah, you're right!

geoffwhittington:
This looks like it should work assuming that `references` is always returned for v4.0 servers in tasks, and I think it does. 

:+1:

sibartlett:
I have checked that `references` is always returned :)

PR #389, opened 2016-08-17
GitHub - implements MacroAlmConnector

sibartlett:
Refs #SDE-3032


sdetools/modules/sync_github/github_plugin.py
```diff
@@ -232,14 +237,9 @@ def alm_add_task(self, task):
...
         labels = []
-        task_content = PUBLIC_TASK_CONTENT
-
-        if not self.sync_titles_only:
-            task_content = self.sde_get_task_content(task)
 
         create_args = {
-            'title': task['alm_full_title'],
-            'body': task_content,
+            'title': task['alm_full_title']
```
geoffwhittington: I forget, why do we do this? Why can't everything be in `GITHUB_DEFAULT_ISSUE_TEMPLATE`?

geoffwhittington: Oh I know why, we don't have a macro for the value of `task['alm_full_title']`. If we had that, which is just `self.get_alm_task_title(self.config, task, fixed=False)` from `AlmConnector` then we could do it.

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -1054,6 +1054,7 @@ def transform_task(self, task, project=None):
             'id': task['id'],
             'task_id': task_id,
             'task_title': re.sub('^%s:\s+' % task_id, '', task['title']),
+            'task_alm_full_title': task['alm_full_title'],
```
sibartlett: @geoffwhittington do you think "task_alm_full_title" makes a good macro name? Or you got any better ideas?

geoffwhittington:
Bold! merged before CI status was green

PR #394, opened 2016-08-19
GitHub - task references

sibartlett:


sdetools/modules/sync_github/github_plugin.py
```diff
@@ -79,13 +79,18 @@ def parse_error(self, result):
...
         self.task_id = task_id
         self.alm_id = alm_id
         self.status = status
         self.timestamp = timestamp
         self.priority = None
 
+        # Task Reference Info
+        self.link = link
+        self.reference = alm_id
+        self.name = alm_id
```
sibartlett: Given that GitHub issues just have a numerical ID; am I correct to assume we want task reference `name` to be equal to `alm_id`?

@shawngustaw @geoffwhittington

geoffwhittington: yep sounds right

PR #396, opened 2016-08-22
RTC - task references

sibartlett:


sdetools/modules/sync_rational/rational_plugin.py
```diff
@@ -381,7 +387,23 @@ def alm_validate_configurations(self):
         if invalid_fields:
             raise AlmException('Invalid field(s) specified: %s' % ', '.join(invalid_fields))
 
-    def alm_get_task(self, task):
+    def alm_get_task_by_reference(self, task, reference):
+        work_item_url = 'resource/itemName/com.ibm.team.workitem.WorkItem/%s' % reference['reference']
+        try:
+            work_item = self._call_api(work_item_url)
+        except APIHTTPError as e:
+            if e.code == 403:
```
sibartlett: The RTC API returns 403 if a work item has not been created.

I was unable to delete any work items, so was unable to confirm what HTTP code is returned for a deleted work item.

sdetools/modules/sync_rational/rational_plugin.py
```diff
@@ -382,7 +388,23 @@ def alm_validate_configurations(self):
         if invalid_fields:
             raise AlmException('Invalid field(s) specified: %s' % ', '.join(invalid_fields))
 
-    def alm_get_task(self, task):
+    def alm_get_task_by_reference(self, task, reference):
+        work_item_url = 'resource/itemName/com.ibm.team.workitem.WorkItem/%s' % reference['reference']
```
geoffwhittington: I don't know if is always true? With RTC you can make Story and other issue types. Is the URL different with these types or are they all work items?

geoffwhittington: You can get the full URL of the issue from the API, I don't think we should be worried about storing longer references (unless of course the field is too restricted)

sibartlett:
@geoffwhittington I changed the reference to be the work item's URL.

geoffwhittington:
cool !

PR #398, opened 2016-08-23
Convert Mingle to MacroAlmConnector

MartinPetkov:
Refs #SDE-3033


sdetools/modules/sync_mingle/mingle_plugin.py
```diff
@@ -9,19 +9,31 @@
...
 logger = log_mgr.mods.add_mod(__name__)
 
+MINGLE_DEFAULT_PRIORITY_MAP = {
+    '7-10': 'High',
+    '4-6': 'Medium',
+    '1-3': 'Low',
+}
+
+MINGLE_DEFAULT_ISSUE_TEMPLATE = {
+    'card[name]': '${task_alm_title}',
```
geoffwhittington: @sibartlett I suggested to @MartinPetkov that we want these keys to be the human-readable field Titles from the Mingle UI. I know this is part of an on-going discussion but I wasn't aware of any solid decisions, I figured we should do what we're already doing with the other integrations.

sibartlett:
Let's discuss next week (Wednesday perhaps).

geoffwhittington:
@MartinPetkov might as well put this PR on hold for now

geoffwhittington:
By the tests fail because AppVeyor couldn't find a dependency. We should fix the AppVeyor setup.

PR #399, opened 2016-08-25
Update Jira links so they point to a user friendly link

shawngustaw:
Previously they pointed to the api endpoint which isn't
very helpful for the end user.

Refs #SDE-2931


shawngustaw:
@geoffwhittington I tried to do this a little bit nicer but the user friendly links are pretty tightly coupled to the connection and I didn't like passing around the method, server and context root all over the place. Do you think this is fine or would you prefer I refactor that stuff?

sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -194,9 +194,8 @@ def create_task_obj(self, sde_id, jtask):
                         timestamp,
                         self.config['jira_done_statuses'],
                         task_versions,
-                        jtask['self'],
-                        jtask['id'],
-                        jtask['key']
+                        reference=jtask['id'],
```
geoffwhittington: Would be good to use either positional or named arguments. I'm not sure by looking at things what is going to happen here.

geoffwhittington:
What would you do differently? I don't see what you could do better.

shawngustaw:
Haha in a perfect world things would all just happen in one place instead of doing the patching after returning the task but it gets the job done.

PR #400, opened 2016-08-30
Fix RTC timeout issue

sibartlett:
Use a web interface endpoint to load enumerated values.


sibartlett:
Closing - as I don't think this is a viable solution, especially for fields that have dependencies.

PR #402, opened 2016-08-31
Use pypi

geoffwhittington:


geoffwhittington:
@sibartlett this should fix the random failures. The commit message is a little light tho ;)

PR #403, opened 2016-09-05
Updated versionone to use macros

wisznie:


sdetools/modules/sync_versionone/versionone_plugin.py
```diff
@@ -358,13 +366,13 @@ def alm_get_task(self, task):
     def alm_add_task(self, task):
         # Create a dictionary of field->value pairs we are ready to set on the asset
         new_asset_fields = {
-            'Name': task['alm_full_title'].encode('ascii', 'xmlcharrefreplace'),
-            'Description': self.sde_get_task_content(task).encode('ascii', 'xmlcharrefreplace'),
-            'Priority': self.translate_priority(task['priority']),
-            'Project': self.config['alm_project'],
-            'Status': self.config['v1_new_status'],
+            'Status': self.config['v1_new_status']
```
wisznie: This is the one field I didn't macrofy since it seems to be specific to v1.

sdetools/modules/sync_versionone/versionone_plugin.py
```diff
@@ -358,13 +366,13 @@ def alm_get_task(self, task):
     def alm_add_task(self, task):
         # Create a dictionary of field->value pairs we are ready to set on the asset
         new_asset_fields = {
-            'Name': task['alm_full_title'].encode('ascii', 'xmlcharrefreplace'),
```
geoffwhittington: We ran into some bugs by not encoding this and the line below. So I think the solution has to incorporate that, .... somehow.

wisznie: Why not just do the encoding again in here? The value is still derived from the macro. I think that's the best way and the updated diff reflects that.

PR #407, opened 2016-09-07
Feature/tfs task refs

keyvanm:
I think I accidentally merged in the RTC task references branch too. If that causes problems please let me know so I can clean it up.


keyvanm:
if I change the title of a task in TFS (or Jira) and sync again, the old task gets recreated and then I get this error:
`APICallError: HTTP Error 400. Explanation returned: FATAL ERROR: {"errors":["Task reference with this Task and Alm connection already exists."]}`

Error happens in TFS and Jira (the ones I tested).

sibartlett:
The HTTP 400 error you described, is definitely a bug.

sdetools should be using the task reference information to load the JIRA or TFS task - even if their title has been changed.

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -163,8 +179,9 @@ def alm_validate_configurations(self):
             raise AlmException('Invalid field(s) specified: {0}'.format(', '.join(invalid_fields)))
 
     def alm_get_task(self, task):
```
sibartlett: `alm_get_task` should be renamed to `alm_get_task_legacy`.

It also should not call `alm_get_task_by_reference`, you can remove the first 3 lines of the method.

The base class' `alm_get_task` method determines which of the 2 methods to call.

MartinPetkov:
@keyvanm @sibartlett I could not reproduce the HTTP 400 bug, as it was explained. These are the steps I took:
1. Ran the initial sync between SDE and Jira
2. Went on Jira and changed the title of one of the tasks that was synced
3. Ran the sync again

I did not encounter any bugs.

sibartlett:
I've cleaned this branch up, and opened a new merge request here: #412

PR #410, opened 2016-09-09
get task by reference versioneone v1

wisznie:


sdetools/modules/sync_versionone/versionone_plugin.py
```diff
@@ -327,7 +334,40 @@ def alm_connect_project(self):
         if not self.project:
             raise AlmException('Project %s could not be found' % self.config['alm_project'])
 
-    def alm_get_task(self, task):
+    def alm_get_task_by_reference(self, task, reference):
```
sibartlett: I am not familiar with the VersionOne API, but is there no way to do just 1 request instead of 2?

wisznie: The only way to get a link to the task is by making the second call, that link is easily constructible with the information known after the first call, it's just "/<task_type>/<_oid>", but the second call gives the link that is guaranteed to be the link should versionone change in the future. I'll take a look through the api again to see I didn't miss some alternative.

So in short, we can just construct the url for the link and do one api call(and leave a comment in case there's a breaking change in the api in the future to do the second call) or we can do both calls.

wisznie: Following up on this. We can just construct the url link and not do the second api call. I can make that change and update the PR tonight to reflect that if it that is preferred.

sibartlett: I'm thinking we should keep the 2nd API call - to mitigate the risk of a breaking change. If it's an issue, we can always revert later.

sibartlett:
Merged in #416

PR #415, opened 2016-09-16
Rally - get task object from create response

Gee19:
Refs #SDE-3143


Gee19:
TODO:
- update mock test response data

PR #416, opened 2016-09-17
VersionOne - task references

sibartlett:
Refs #SDE-3040


geoffwhittington:
I don't see any issues with this PR when I test against v4.0 (m1).

shawngustaw:
Taking a look at this now but make sure you squash before you merge.

sibartlett:
This pull request needs to be opened against develop too.

sdetools/modules/sync_versionone/versionone_plugin.py
```diff
@@ -79,14 +79,19 @@ def parse_response(self, result, headers):
...
         self.asset = asset
         self.done_statuses = done_statuses  # comma-separated list
         self.priority = None
         self.status = asset['status']
         self.timestamp = asset['ChangeDateUTC']
 
+        # Task Reference Info
+        self.link = link
+        self.reference = self.asset['Number']
+        self.name = self.asset['Name']
```
shawngustaw: I think this should be set to asset['_oid'], otherwise this is always going to be set to the name of the SDETask.

In Jira we follow this structure:
link = link a user can click on to view the task in Jira (not a rest api link)
reference = internal unique ID you can query rest API directly to retrieve it
name = name that Jira gives to tasks, i.e. in our Jira instance tasks are titled #SDE-1234

sibartlett: I think we should prefer `self.asset['Number']` over `self.asset['_oid']`

shawngustaw: I know we do the same for the Github integration but it doesn't have any sort of "user-friendly" name.

geoffwhittington: The user-friendly name I thought we'd use with V1 is the ID field as in the screenshot
v1_id

sibartlett: @geoffwhittington that's what we switched to

geoffwhittington: :+1:

PR #422, opened 2016-09-26
More descriptive error messages containing status code

shawngustaw:


sdetools/sdelib/restclient.py
```diff
@@ -351,7 +351,10 @@ def call_api(self, target, method=URLRequest.GET, args=None, call_headers={}, au
                 raise ServerError('Invalid server or server unreachable: %s' % (self.server))
             try:
                 err_msg = handle.read()
-                logger.info('Error calling %s API. Raw error: %s' % (self.conf_name, repr(err_msg)[:200]))
+                log_msg = 'Error %s calling %s API.' % (handle.code, self.conf_name)
```
sibartlett: Do we always expect the error to have a `code` attribute defined? What if it's not a HTTP error, is that possible?

sibartlett: I guess `code` must always be defined, because at the end of the method we have this:

```
if handle.code == 401:
    raise APIAuthError('Invalid Credentials for %s' % self.conf_name)
raise APICallError(handle.code, err_msg[:255])
```

shawngustaw: @sibartlett there's certain cases where it might return None, but at the end of this code block we do

```if handle.code == 401:```

which leads me to believe we haven't hit the case where urllib2.open returns None

shawngustaw: (forgot to post documentation:

https://docs.python.org/2/library/urllib2.html#urllib2.urlopen)

shawngustaw: Clearly hadn't refreshed before I responded haha...

PR #427, opened 2016-10-17
Listen for program abort and cleanup cert file

geoffwhittington:
Trying to use https://github.com/antevens/listen/ to catch program abort to properly cleanup certs

@antevens I'm doing this wrong, right? :) I can't seem to trigger `remove_file`


geoffwhittington:
Interesting this works if I trigger a ctrl-c but not if I let the python process end  normally. I must be missing a signal type

geoffwhittington:
@sibartlett The cert file can now be cleaned up during normal and abnormal termination

PR #432, opened 2016-11-23
ALM refactor base class & plugin PoC's

Gee19:


sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,374 @@
...
+                IntegrationField(
+                    name='conflict_policy',
+                    label='Choose Authoritative Source',
+                    field_type=FieldType.RADIO,
+                    cli_description='Conflict policy to use',
+                    default='alm',
+                    options=[
+                        Option('ALM', 'alm'),
+                        Option('SD Elements', 'sde'),
+                        Option('Last Status Change', 'timestamp', disabled=True)
```
geoffwhittington: Note "Last Status Change" only applies to Jira.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,374 @@
...
+                cli_description='Standard workflow in ALM?',
+                default='True'
+            ),
+            IntegrationField(
+                name='selected_tasks',
+                field_type=FieldType.STRING,
+                cli_description=("Optionally limit the sync to certain tasks "
+                                 "(comma separated, e.g. T12,T13). Note: Overrides other selections.")
+            ),
+            IntegrationField(
```
geoffwhittington: Note: this is a command-line-only relevant option. We could even remove it completely / not support it in-app.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,374 @@
...
+                name='show_progress',
+                field_type=FieldType.BOOLEAN,
+                cli_description='Show progress',
+                default='False'
+            ),
+            IntegrationField(
+                name='test_alm',
```
geoffwhittington: This one is a bit odd, I don't expect we'd expose this as a selectable option - if you look at the code what it does it short-circuit the integration so it tests one of the three options.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,374 @@
...
+                field_type=FieldType.STRING,
+                cli_description='Test Alm "server", "project" or "settings" configuration only'
+            ),
+            IntegrationField(
+                name='alm_custom_fields',
+                field_type=FieldType.STRING,
```
geoffwhittington: Technically this is a string, but JSON is more apt.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,374 @@
...
+                cli_description=("Customized fields to include when creating a task in %s "
+                                 "(JSON encoded dictionary of strings)") % self.name
+            ),
+            IntegrationField(
+                name='alm_reference_context',
+                field_type=FieldType.STRING,
```
geoffwhittington: I'm not sure about this one, I don't think you'd show this on the front-end.

sdetools/alm_integration/github_integration_plugin.py
```diff
@@ -0,0 +1,87 @@
+from alm_integration_plugin_base import (BaseALMIntegration,
+                                         FieldType, Option, Section, IntegrationField)
+
+
+class GithubIntegrationPlugin(BaseALMIntegration):
+
+    TITLE_FORMAT_OPTIONS = [
```
geoffwhittington: These can be moved to the base class and shared by all ALMs

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,384 @@
...
+                    ]
+                )
+            ])
+
+    def get_sections(self):
+        """
+        Return a list of fields pertaining to the integration, SDE will
+        dynamically generate these on the form.
+        """
+        return [
```
hanifvirani: You don't include `advanced_options_sections` here but if I look at the `get_sections` of the Github connector, I see that you append `advanced_options_sections` there. Is that by choice?

hanifvirani: I think it might be good for only the base class to have the `sections` property which composes various sections together which the child classes override.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: Yes, this is why it is disabled by default, and we enable that field only for JIRA.

sdetools/alm_integration/alm_integration_plugin_base.py
Gee19: The reasoning for this is that in the future this plugin will be used throughout SDE Tools.

We are also not returning this in the API as it is only part of get_cli_only_fields.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: As stated above - these plugins are meant to contain all fields (not just the ones in the ui)

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,382 @@
...
+                    options=[
+                        Option('No Status', 'none'),
+                        Option('Partial Pass', 'partial'),
+                        Option('Pass', 'pass'),
+                        Option('Fail', 'fail')
+                    ]
+                )
+            ])
+
+    def get_sections(self):
```
lmontopo: As discussed, I like it the way I have it better.  We can discuss / decide more on this later, though.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: Our discussion yesterday revealed the following: 
- if possible: it would be good for the BaseALMIntegration to define the 'get_sections' method, and have that *never* be overridden by the subclasses.  If possible, it would return the correct output for each subclass without having to be overwritten. 
- We are going to put thought into how this may be achieved.

lmontopo:
@Gee19 I don't think I have permissions to edit this MR since you opened it, but ideally it should be between this branch and the 'feature/sdetools_refactor' branch.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -147,7 +147,8 @@ def _add_config_options(self):
         self.config.opts.add("group_query",
                              "LDAP query to get all groups.")
 
-    def normalize_dc(self, dc_section):
+    @staticmethod
+    def normalize_dc(dc_section):
```
lmontopo: I rebased sdetools_refactor off of master, and then this branch off of sdetools_refactor, but these differences still appear here?

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+class Option(object):
+
+    def __init__(self, label, value, disabled=False):
+        self.label = label
+        self.value = value
+        self.disabled = disabled
+
+
+class Section(object):
+
```
lmontopo: docstring - Simon

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+        self.fields = fields
+
+
+class IntegrationField(object):
+    """
+    Used by the ALM integrations to describe a field.
+    """
+
+    def __init__(self, name, label='', description='', cli_description='', options=None, help_text="", max_length=None,
+                 field_type=FieldType.STRING, disabled=False, default=None, required=False, placeholder=None):
```
lmontopo: Add Docstring (or comment) that describes what all the fields all (they aren't all self explanatory) - Simon.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+        self.description = description
+        self.cli_description = cli_description
+        self.max_length = max_length
+        self.required = required
+        self.placeholder = placeholder
+        self.options = options or []
+
+    def run_validation(self, value, exclude_required_validation=False):
+        if not exclude_required_validation and self.required:
+            if not value:
```
lmontopo: combine these two ifs into one line. - Simon

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+                raise IntegrationPluginException('Field {0} is required.'.format(self.label))
+
+        if not value:
+            return
+
+        if self.max_length:
+            if len(value) > self.max_length:
```
lmontopo: combine ifs. - simon.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: Also could be a good idea to document some cases (like when would disabled = True? what do we mean?)

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+                raise IntegrationPluginException('Max length of field {0} is {1}.'.format(self.label, self.max_length))
+
+        if self.disabled:
+            raise IntegrationPluginException('The following field is disabled: {0}'.format(self.name))
```
lmontopo: this logic is currently wrong.  

if it has a default - make it check that the value is set to default, otherwise raise error.  Don't always raise error. 

Also - maybe `readonly` is a more appropriate name?

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+
+        if len(self.options) > 0:
```
lmontopo: `if self.options` is better.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,374 @@
...
+        Option('7', 7),
+        Option('6', 6),
+        Option('5 - Medium Minimum Priority to Import', 5),
+        Option('4', 4),
+        Option('3', 3),
+        Option('2', 2),
+        Option('1 - Low Minimum Priority to Import', 1)
+    ]
+
+    DYNAMIC_MACROS = ("Field values can use the following dynamic macros. Their values change depending on the "
```
lmontopo: OK we will no longer include it in the description.  More discussion is needed regarding help text. - Simon and Leta and Jordan

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+
+    @abstractproperty
+    def command(self):
+        """ Return a command name used for syncing the specific integration """
+        pass
+
+    def get_alm_project_field(self):
+        # can't get rid of this yet because other alm's haven't transitioned
+        # to new style - hence they just overwrite this method.
+        return self.alm_project
```
lmontopo: This method can be gotten rid of I realize now - Leta

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+
+    def get_advanced_options_section(self):
+        return Section(
+            id='advanced_options',
+            label='Advanced ALM Options',
+            fields=[self.alm_context]
+        )
```
lmontopo: Idea - make helper methods that would implement either token or username password.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+
+    @abstractmethod
+    def get_creds_section(self):
+        """
+        Returns a Section object with the credential fields.
+        Each ALM implementation must overwrite this method.
+        """
+        pass
```
lmontopo: We should implement helper methods for token or username password.  This will still be an abstract mehtod. 

- simon.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+
+    def get_connection_details_section(self):
+        return Section(
+            id='connection_details',
+            label='Connection Details',
+            fields=[
+                self.alm_method,
+                self.server,
```
lmontopo: we should default this field on Github to api.github.com.  Just do it. - Simon.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+            ])
+
+    def get_sections(self):
+        """
+        Return a list of fields pertaining to the integration, SDE will
+        dynamically generate these on the form.
+        """
+        return [
+            self.get_connection_details_section(),
+            self.get_creds_section(),
```
lmontopo: rename this to get_credentials_section

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+            self.start_fresh,
+            self.alm_standard_workflow,
+            self.selected_tasks,
+            self.show_progress,
+            self.test_alm,
+            self.alm_custom_fields,
+            self.alm_reference_context
+        ]
+
+    def insert_after(self, the_list, after, items_to_insert):
```
lmontopo: lets get rid of this for now and put it in a snippet or gist since we aren't using it anymore (but maybe might need to use it in the future).

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+
+        for item in reversed(items_to_insert):
+            try:
+                the_list.insert(the_list.index(after_obj) + 1, item)
+            except ValueError as exc:
+                raise IntegrationPluginException("Item doesn't exist in the list: {0}".format(exc.message))
+
+    def get_field(self, name, include_cli=False):
+        fields = self.get_all_fields(include_cli)
+        return next([field for field in fields if field.name == name])
```
lmontopo: maybe same? get rid of? - Leta, Simon, Jordan.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,422 @@
...
+
+        for section in self.get_sections():
+            fields.extend(section.fields)
+
+        fields.append(self.get_alm_project_field())
+
+        return fields
+
+    def validate(self, params, include_cli=False, exclude_required_validation=None):
+        errors = {}
```
lmontopo: we should copy params since we're modifying it.

sdetools/alm_integration/github_integration_plugin.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @property
+    def command(self):
+        return 'sync_github'
+
+    def __init__(self):
+        super(GithubIntegrationPlugin, self).__init__()
+
+        #  update alm_project field (new method)
+        self.alm_project.label = 'ALM Project:'
```
sibartlett: Why not have `ALM Project` as the default label for this field in the base class?

sibartlett: I think we should also drop the colon from all label strings. If the form fields need colons, then SDE should be adding those it itself.

lmontopo: cool, i like that better too actually.

lmontopo: good idea.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: done.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: name is changed to `readonly` and logic is fixed.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: done.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: done.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: just put it in a gist.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: gist created.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: done.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: done.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: done.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: done.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: done.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: I forgot that we actually are using this still for ordering sections.  We no longer are inserting fields midway into a section, but we are inserting a section after another section, so i had to put this method back in for our JIRA plugin to work.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -33,17 +41,62 @@ def __init__(self, id, label, fields):
...
+                 options=None, help_text="", max_length=None,
+                 field_type=FieldType.STRING, readonly=False, default=None,
+                 required=False, placeholder=None):
+        """
+        :param name: the unique identifier of this field
+
+        :param label: the label of this field in the UI
+
+        :param description: the description of this field in the UI, for example
+                            the description of the field lebelled 'github repo
```
Gee19: typo

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -33,17 +41,62 @@ def __init__(self, id, label, fields):
...
+                            owner' would be could be 'The owner of the github
```
Gee19: wording (would be could be)

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,514 @@
...
+                Option('Development', 'development'),
+                Option('Testing', 'testing')
+            ],
+        )
+        self.sde_tags_filter = IntegrationField(
+            name='sde_tags_filter',
+            label="Tasks having all of the following tags (optional)",
+            field_type=FieldType.STRING,
+            placeholder='New Tag...',
+            help_text='Filter Project Tasks by tag (tag1, tag2)'
```
sibartlett: This one should be `description`.

Gee19: Fixed.

PR #435, opened 2016-12-07
Expose LDAP Helper methods as class methods

RouganStriker:
This will allow us to re-use some of the logic inside SDE


emintham:
+1

PR #437, opened 2016-12-14
Add HSTS to AppScan mapping

geoffwhittington:


geoffwhittington:
@sibartlett can you review and merge if good?

PR #441, opened 2017-01-09
Add Checkbox Feature

lmontopo:
Refs: #SDE-3518


lmontopo:
Does it suffice to just have each field tell us whether or not it is 'editable_on_project' by default, or should we implement a method on the plugin that returns a list of field names which correspond to the fields which are editable on the project by default? 

I guess its either we do the work *here* to search through all fields and filter down to the ones that are editable, or we do similar lookup or filetering work in the frontend when we need this info.

sibartlett:
+1

PR #443, opened 2017-01-11
AppScan Mobile Analyzer report support

geoffwhittington:


sibartlett:
Wow, you're quick!

geoffwhittington:
I verified this by running against the XML we have and looking at the findings it picks up. I think I parsed the file correctly...

geoffwhittington:
I've realized that this MR is not correct, we can't make decisions on task verification using a mapping file that includes AppScan Standard, Enterprise and Mobile Analyzer. We need to make sure we only load the task > weakness mapping that corresponds to Appscan Mobile Analyzer. So there are two ways to do this, either write a net new import_appscan_mobile_analyzer sdetools command or we do the following:

[ ] Create a new mapping file for AppScan Mobile Analyzer
[ ] Add a check to https://github.com/sdelements/sdetools/blob/master/sdetools/modules/import_appscan/appscan_integrator.py#L109 for `mobile_analyzer` and make sure we load the mobile analyer mapping only.

The main app should also be updated so that "mobile_analyzer" corresponds to a legit scanner.

PR #445, opened 2017-01-18
ALM Integration Plugin Base validation tests

Gee19:


sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_base_exclude_required_validation(self):
+        params = {'alm_project': 'heyo'}
+        exclude = [self.mock.test_field.name]
+
+        self.mock.validate(params=params, exclude_required_validation=exclude)
+
+    def test_field_exclude_required_validation_raises_no_errors(self):
+        self.mock.test_field.run_validation(value='', exclude_required_validation=True)
+
+    def test_base_validation_returns_all_errors(self):
```
lmontopo: I would change the name here to be `test_plugin_validation_returns_all_errors`.  In general i think the name `plugin` is better than `base` in these test names.  It took me a bit to understand what you meant by `base` but I think you're just trying to specify that this test is testing the plugin validation as a whole as opposed to just the field level validation.

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
```diff
@@ -0,0 +1,114 @@
...
+    DISABLED_ERROR = 'The following field is disabled: readonly_field'
+    BASE_VALIDATE_ERRORS = [
+        'Field ALM Project is required.',
+        'Max length of field Test Field is 5.'
+    ]
+
+    def setUp(self):
+        self.mock = MockBaseALMIntegration()
+
+    def test_base_exclude_required_validation(self):
```
lmontopo: maybe put a comment here to say that we're just ensuring no error is raised (otherwise I think its confusing as to why we have no assert statement.  Maybe @simon has further advice on this?  Is it OK in general not to have any assert statements given that all we want to do is ensure no error is raised?

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
```diff
@@ -0,0 +1,114 @@
+import unittest
+
+from sdetools.alm_integration.alm_integration_plugin_base import BaseALMIntegration, IntegrationField, FieldType, Option, Section
+
+from sdetools.sde_plugins.exceptions import PluginException
+
+
+class ALMIntegrationPluginBaseTests(unittest.TestCase):
+
+    # note: we should stay consistent when returning field.name or field.label in our error msgs
```
lmontopo: Good point.  Lets do this.

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
```diff
@@ -0,0 +1,114 @@
...
+    REQUIRED_ERROR = 'Field Test Field is required.'
+    MAX_LENGTH_ERROR = 'Max length of field Test Field is 5.'
+    OPTION_ERROR = 'almyo is an invalid option for field conflict_policy'
+    DISABLED_ERROR = 'The following field is disabled: readonly_field'
```
lmontopo: I actually think that we should change this error to be "The following field is readonly".  I think I just missed this when I changed the `IntegrationField` attribute itself to be `readonly` instead of `disabled`.  Thoughts?

(Its good that writing these tests is making us aware of some inconsistencies!)

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
Gee19: I think it is better to let the test fail if an error is raised and assume a pass otherwise, it makes the code much more readable. Although, I am open to suggestions :)

Discussion here:
http://stackoverflow.com/questions/4319825/python-unittest-opposite-of-assertraises

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
Gee19: I agree, my naming was pretty bad.

Fixed.

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
lmontopo: OK good point.  I still think a short description might be useful.

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
Gee19: Switched all the errors to field.name.

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
Gee19: Fixed.

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
Gee19: I renamed the test to be more descriptive, let me know what you think.

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
```diff
@@ -0,0 +1,118 @@
...
+
+class ALMIntegrationPluginBaseTests(unittest.TestCase):
+    REQUIRED_ERROR = 'Field test_field is required.'
+    MAX_LENGTH_ERROR = 'Max length of field test_field is 5.'
+    OPTION_ERROR = 'almyo is an invalid option for field conflict_policy.'
+    READONLY_ERROR = 'Field readonly_field is readonly.'
+    BASE_VALIDATE_ERRORS = [
+        'Field alm_project is required.',
+        'Max length of field test_field is 5.'
+    ]
```
lmontopo: Now that I'm thinking about it, it probably makes more sense to have some of these errors (like MAX_LENGTH_ERROR, REQUIRED_ERROR, etc) to be moved to the IntegraitonField definition in a nested class called Messages, the way we do in our main Django app.  Then we woudln't have to redefine the messages here, we could just refer to them and pass in whatever strings we needed to pass in you know?  

That doesn't necessarily have to be done in your MR since its a separate issue, but I just thought i'd mention it.

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def setUp(self):
+        self.mock = MockBaseALMIntegration()
+
+    def test_plugin_exclude_required_validation_raises_no_errors(self):
```
lmontopo: nice.  perfect name.

sdetools/alm_integration/tests/alm_integration_plugin_test_base.py
```diff
@@ -0,0 +1,118 @@
...
+    def test_plugin_validation_returns_all_errors(self):
+        params = {'test_field': '123456'}
+
+        with self.assertRaises(PluginException) as e:
+            self.mock.validate(params)
+
+        self.assertEqual(e.exception.messages, self.BASE_VALIDATE_ERRORS)
+
+    def test_plugin_validation_removes_garbage_fields(self):
+        params = {'garbage_fields': 'will_be_removed', 'test_field': 'stays'}
```
lmontopo: i like this!!

lmontopo:
This looks really good to me and I think you've hit all the 'validation' test cases I can think of.  

One thought I have while reading this is wether or not we should split up these tests into two separate test classes or not.  I wonder if it makes sense to keep all of the field validation tests in an `IntegrationFieldTest` class instead of in the `ALMIntegrationPluginBaseTests`?  Just a thought.  If we did do that, perhaps we'd not instantiate the mock ALM each time, but just instantiate the `IntegrationField`'s that we need?  I think this would make sense going forward, too, since eventually we'd probably add other tests (for things other than the validation) to our `ALMIntegrationPluginBaseTests`.  Let me know what you think.  This isn't a blocker by any means - so it can also always be done later if we feel its a good move.

lmontopo:
+1 with or without any of the suggestions (separating into 2 test classes and making a messages class inside IntegrationField).  Good work, I think you learned lots doing this work which is awesome.

lmontopo:
@Gee19 I figured out why some of your tests are failing.  The ability to use `assertRaises` as a context manager was added to python 2.7.  So in python 2.6 this doesn't work.  Since Tools needs to be compatible with python 2.6, we need to update these tests to use `assertRaises` but NOT as a context manager. I am reading about it, and it appears like its difficult to make assertions about the error message without using the context manager version, so, actually, people are saying we can 'backport' the 2.7 version of unittest and use it in python 2.6.  (I'm reading here: http://stackoverflow.com/questions/129507/how-do-you-test-that-a-python-function-throws-an-exception) So i'm in favour of doing that.  See answer 52: `import unittest2 as unittest`.

lmontopo:
Once the builds pass, I re-instate my +1. :)

aeldamaty:
After getting rid of those methods if you can, +1!

PR #446, opened 2017-01-18
Feature/3578 rally plugin

lmontopo:
Implement CAAgileCentralIntegrationPlugin.


sdetools/alm_integration/ca_agile_central_integration_plugin.py
```diff
@@ -0,0 +1,80 @@
...
+        self.rally_new_statuses = IntegrationField(
+            name='rally_new_status',
+            label='Open an issue in CA Agile Central with the following state',
+            cli_description='status to set for new tasks in Rally',
+            placeholder='Defined'
+        )
+        self.workspace = IntegrationField(
+            name='rally_workspace',
+            label='CA Agile Central Workspace',
+            placeholder='My Workspace'
```
geoffwhittington: I wouldn't put a placeholder for this as its required and no good default can be provided.

sibartlett: Placeholder != default

A placeholder is example input. Although, I'm not sure "My Workspace" is a useful example?

lmontopo: That is the current placeholder on the form, which is why I put it there.  I'm open to removing it and/or putting something more useful there. 

Thanks for reminding me that this is a required field!  I will add that!

geoffwhittington: Usually we set Placeholders for the default value we have in sdetools commands. I just assumed it was communicating the same thing.

I've seen IE do terrible things in the past with Placeholders - making them the default value.

lmontopo:
Question @sibartlett @geoffwhittington : Its wasn't clear to me which things should reference 'rally' and which things should reference 'ca agile central'.  I kept the UI things consistent with what was already there (so ca agile central) and i left the id 'rally' because thats our current id.  But like, the file name for instance? And just variables in the code.  Should they consistently be 'rally' or should we try to use 'ca agile central' everywhere except for the id? Let me know what your thoughts are on this.

sdetools/alm_integration/ca_agile_central_integration_plugin.py
sibartlett: IE didn't get support for placeholders until version 10.

We've been using some flaky hack to add "placeholder" support to old versions of browsers that don't support them - hence the IE issues you describe.

sibartlett:
I say we use "rally" everywhere that's not visible by user, just because it's nice and short. The UI must always say "CA Agile Central" though, because that's the new product name.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -315,6 +315,82 @@ def __init__(self):
...
+            field_type=FieldType.STRING,
+            max_length=1024,
+            editable_on_project=True,
+            cli_description=(
+                "Customized fields to include when creating a task in {0} "
+                "(JSON encoded dictionary of strings)").format(self.name),
+            description=(
+                "If you have customized your ALM you may need to fill in the "
+                "following field. Please get in touch with SD Elements support"
+                " for help.\n{0}".format(self.DYNAMIC_MACROS))
```
geoffwhittington: We could remove the `{0}` and the `DYNAMIC_MACROS` - we could only show a couple here and they'd need to be in context.

sdetools/alm_integration/ca_agile_central_integration_plugin.py
```diff
@@ -0,0 +1,81 @@
...
+        super(CAAgileCentralIntegrationPlugin, self).__init__()
+
+        # update alm project field
+        self.alm_project.help_text = 'CA Agile Central Help Text Here'
+
+        self.card_type = IntegrationField(
+            name='rally_card_type',
+            label='Issue Type',
+            placeholder='Story',
+            cli_description='IDs for issues raised in Rally'
```
geoffwhittington: Description should read, `The type of issue to create in the CA Agile Central project`

sdetools/alm_integration/ca_agile_central_integration_plugin.py
```diff
@@ -0,0 +1,81 @@
...
+        self.rally_done_statuses = IntegrationField(
+            name='rally_done_statuses',
+            label='States which map to DONE in SD Elements',
+            cli_description='statuses that signify a task is Done in Rally',
+            placeholder='Completed,Accepted'
+        )
+        self.rally_new_statuses = IntegrationField(
+            name='rally_new_status',
+            label='Open an issue in CA Agile Central with the following state',
+            cli_description='status to set for new tasks in Rally',
```
geoffwhittington: `Status to assign to new issues in CA Agile Central`

sdetools/alm_integration/ca_agile_central_integration_plugin.py
lmontopo: OK. The `cli_description`'s that I've put in this document all came from what currently is the text on the command line.  I'm not against updating these strings, but just so you know that's why they are what they currently are.

sdetools/alm_integration/ca_agile_central_integration_plugin.py
lmontopo: Cool.  The command line currently refers to the alm as 'rally' but I can update this to be CA Agile Central.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: not sure i understand what you mean exactly.  Why would we not show all of the options?

lmontopo:
Fixed all the changes except the DYNAMIC MACROS one and rebased so that I'm on top of Jordan's i18n changes, and I internationalized the few strings I added.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: Ok so after talking to Simon I think that you are just referring to maybe not having this huge description here, but instead maybe listing this info in a Zendesk doc or some other way.  This is something we've discussed and that we will consider separately.  It is outside of the scope of this ticket.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -317,6 +317,80 @@ def __init__(self):
             required=True,
             editable_on_project=True
         )
+        self.title_format = IntegrationField(
+            name='title_format',
+            label=_('ALM Title Format'),
+            field_type=FieldType.SELECT,
+            description=_("Customize the issue titles created in the "
+                          " ALM by choosing one of the templates below."),
```
Gee19: extra space here

Gee19:
+1 even though I made some very minor changes to this branch :)

sibartlett:
+1

PR #448, opened 2017-01-31
Add default fields to base report plugin

Gee19:
Fields default to empty, globalBU report will define what
features & permissions are needed.

Refs #SDE-3601


emintham:
+1. We'll also need to release a version of sdetools and bump up the sdetools dependencies in SDE.

PR #449, opened 2017-02-03
Fix bug with error reporting

emintham:
- also adds logging


geoffwhittington:
+1

aeldamaty:
+1

Gee19:
You could also call str where they are defined, either way +1 :)

PR #450, opened 2017-02-06
Fixes Suds exceptions crashing Celery

JeffBain:
* In light of suds.TransportError crashing celery workers upon encountering
      it, commit changes how we handle these types of exceptions. Instead of
      letting them propogate up the call stack, we now catch them and raise our
      own friendly SDEToolsJobError exceptions which don't exhibit issues upon
      being pickled.

Refs SDE-3463


funkaoshi:
@geoffwhittington / @sibartlett this seems fine, yeah?

funkaoshi:
@JeffBain you should also bump the sdetools version number and junk like that. If you look back a few commits you can see what we do.

sdetools/modules/import_checkmarx/checkmarx_integrator.py
```diff
@@ -95,12 +97,20 @@ def login(self):
...
         """
-        self.checkmarx_wsclient = Client(self.web_service_url)
-        credentials = {"User": self.config['analysis_user'], "Pass": self.config['analysis_pass']}
-        # lcid 1033 is English
-        response = self.checkmarx_wsclient.service.Login(applicationCredentials=credentials, lcid=1033)
-        if not response.IsSuccesfull:
-            raise CheckmarxIntegrationError("Login unsuccessful: %s" % response.ErrorMessage)
+        error = None
+        try:
+            self.checkmarx_wsclient = Client(self.web_service_url)
```
funkaoshi: Can this raise an unexpected exception? If not, i'd move the try so it's just around the `Login` call.

JeffBain: The web_service_url call also can raise the same exception, as it pings the same service.

JeffBain:
@funkaoshi Yeah I was going to bump the SDETools version in a separate commit once this MR went in.

sdetools/modules/import_checkmarx/checkmarx_integrator.py
```diff
@@ -3,6 +3,8 @@
 import StringIO
 import time
 
+from suds.transport import TransportError
```
sibartlett: This should be be:

```
from sdetools.extlib.suds.transport import TransportError
```

sdetools/modules/import_fortify/fortify_ssc_importer.py
```diff
@@ -7,6 +7,7 @@
 import socket
 
 from suds.plugin import MessagePlugin
+from suds.transport import TransportError
 from suds.transport.https import HttpAuthenticated
```
sibartlett: Same here. Should be be importing from `extlib`?

```
from sdetools.extlib.suds...
```

JeffBain: Legit. I don't think I realized we had a copy of suds living in extlib. Though we also have imports directly from suds. How does one determine which is correct? Or are the existing imports just also wrong and have fell through the cracks?

sibartlett: I think the existing imports have just fallen through the cracks. All external dependencies in sdetools should be in extlib.

JeffBain: Alright, maybe I'll update them as well so it's clearer in future.

JeffBain: So this doesn't actually work correctly because our extlib copy of suds doesn't import relatively, it does an absolute import from suds. So while I'm importing extlib.suds.transport.TransportError the library is raising suds.transport.TransportError and this doesn't catch it.

I'm not entirely sure how this works, since I don't have suds installed locally as far as I can tell, so I don't quite know what the 2nd source of suds is that the external lib is referencing/ why the way it imports breaks exception mapping, but fixing this is probably out of scope for this fix. It seems like we've already got imports from suds in sdetools, possibly to get around this very problem.

sibartlett: OK fair enough!

JeffBain: This seems to be the issue we're facing: http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-double-import-trap

Basically we'd have to update suds inside extlib to fix their imports in order to avoid having to import suds directly vs. extlib wise.

sibartlett: We should merge this, and revisit this later. Maybe add this to the tech debt backlog.

JeffBain: Sounds good to me.

PR #451, opened 2017-02-10
Bulk import projects and users

geoffwhittington:


geoffwhittington:
@funkaoshi This was contractor task #1 which looks 80% done. Just missing a logging and some error checking. Having some issue I think with business unit membership

geoffwhittington:
@sibartlett my lint-fu is weak, can you see what needs to be fixed?

PR #452, opened 2017-02-10
Fix LDAP sync to run without any user input

JeffBain:
* We extended LDAP sync to take an extra param around deactivating users so
      that it doesn't depend on user input when deactivating stale users

Refs SDE-3505


funkaoshi:
@JeffBain I'm not sure anyone checks Github regularly, so you should post this is waiting review in chat.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -824,6 +827,16 @@ def deactivate_stale_users(self, tracked_users, sd_users):
         """
         Deactivate users in SDE but not in LDAP.
         """
+
+        # Check to see if the decision has been pre-made (such as in the case
+        # of an import started by the app), otherwise falls back to confirming
+        # with user
+        try:
+            deactivate_users = self.config['deactivate_stale_users']
```
funkaoshi: Do you need the try / catch? 

```
deactivate_users = self.config.get('deactivate_stale_users')
confirm_with_user = deactivate_users is None
```

That handles the case `deactivate_stale_users` is not set, or it's set to None.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -848,15 +861,18 @@ def deactivate_stale_users(self, tracked_users, sd_users):
...
-                        "\t%s\n"
-                        "Confirm? [y/n]",
-                        ', '.join(
-                            [u['email'] for u in users_to_deactivate]
-                        ))
+            if confirm_with_user:
+                logger.info("Deactivating the following users:\n"
+                            "\t%s\n"
+                            "Confirm? [y/n]",
+                            ', '.join(
```
emintham: Might want to change this so that it's joined by newline instead and also sorted.

emintham:
+1

sdetools/modules/sync_ldap/ldap_module.py
JeffBain: Fair point. the config isn't a dict so I didn't think to use .get(), but it seems like the config class still defines get and this covers that usecase nicely.

sdetools/modules/sync_ldap/ldap_module.py
JeffBain: I'll add sorting, but I think for now we probably don't want to change the existing behaviour too much. If there's a lot of names, it might eat up all the space in a given console without a nice way to consume the names (like less) since it's interactive, and if there aren't a lot of names it's still relatively readable with comma separated.

emintham:
+1

PR #454, opened 2017-03-08
Allow LDAP Sync through command_driver

geoffwhittington:


geoffwhittington:
@emintham can you look at this? People should be able to run ldap sync (assuming they `pip install -r requirements`) via

`python sde.py command_driver --sde_api_token=XXXXXX@server --filter_connection_access=all --filter_connection_type=ldap -d`

Does this work for you? I'm assuming you might test locally against develop or richmond.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -180,7 +184,12 @@ def initialize(self):
 
         logger.debug("Initializing LDAP connection")
 
+        self.config.process_boolean_config("report_failures")
+
         self.config.process_json_dict("ldap_filter")
+        for filter in ['users', 'groups']:
```
geoffwhittington: Had to do this because the API endpoint will send the filter with users and groups

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -287,6 +299,11 @@ def search(self, query, params, base=None, scope=None):
             size=self.page_size, cookie=''
         )
 
+        # You have to pass solely raw strings to python-ldap
```
geoffwhittington: This addresses the error `TypeError: ('expected string in list', u'sn')`

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -140,6 +141,9 @@ def _add_config_options(self):
         self.config.opts.add("page_size",
                              "Number of results per page returned from LDAP",
                              default="1000")
+        self.config.opts.add("report_failures",
```
geoffwhittington: I didn't want to create a file every time we synced

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -800,26 +816,33 @@ def sync_intersection(self, intersection, sd_users, directory_users):
         """
         deactivated_user_count = 0
         updated_user_count = 0
+        update_failures = []
 
-        with open(self.bad_users_filename, 'wb') as f:
```
geoffwhittington: Have mercy on me, I think I made these changes correctly.

geoffwhittington:
Remote Integration Client can now do remote LDAP Sync (assuming I change the config to include `filter_connection_access=all`

install/setup.iss
```diff
@@ -73,6 +73,7 @@ var
    APITokenPage: TInputQueryWizardPage;
    ProxyPage: TInputQueryWizardPage;
    CertificatePage: TInputOptionWizardPage;
+   ScheduleJobsPage: TInputOptionWizardPage;
```
emintham: I have no experience with this type of files. Does anyone else have experience with this?

sdetools/extlib/listen/signal_handler.py
```diff
@@ -33,12 +33,21 @@
 
 class SignalHandler(object):
     """ Asynchronous signal handling for OS signals, see 'man kill'"""
-    restart_signals = (signal.SIGHUP, )
```
emintham: Minor style thing, I'd keep these as defaults up here and get rid of the else block below.

sdetools/modules/sync_ldap/ldap_module.py
emintham: Why not just replace the config with whatever the API send back?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -287,6 +299,11 @@ def search(self, query, params, base=None, scope=None):
             size=self.page_size, cookie=''
         )
 
+        # Raw strings needed for python-ldap
+        query = query.encode('utf-8') if query else query
```
emintham: What is query's value when it is falsy?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -800,26 +816,33 @@ def sync_intersection(self, intersection, sd_users, directory_users):
...
+                logger.error(error_message, user)
+
+                update_failures.append([
+                    error_message % user,
+                    str(sd_user),
+                    str(dir_user),
+                ])
+
+        if self.config["report_failures"]:
+            with open(self.update_failures_filename, 'wb') as f:
```
emintham: Hmmm there's actually a bug here. This will overwrite the file if it exists.. Should be 'ab' instead to append. Same with the one below.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -836,58 +859,66 @@ def deactivate_stale_users(self, tracked_users, sd_users):
...
+                    str(user),
+                    '',
+                ])
 
                 error_msg = ("SD Elements user '%s' was not found in the "
                              "directory and is now disabled")
                 logger.debug(error_msg, sd_user["email"])
 
+        if self.config["report_failures"]:
+            with open(self.update_failures_filename, 'wb') as f:
```
emintham: 'ab' here as well.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -896,34 +927,39 @@ def sync_new_users_from_ldap(self, untracked_users, directory_users):
...
+                logger.error(error_message, user)
+
+                update_failures.append([
+                    error_message % user,
+                    user,
+                    str(dir_user),
+                ])
+
+        if self.config["report_failures"]:
+            with open(self.update_failures_filename, 'wb') as f:
```
emintham: 'ab' here as well.

emintham:
Someone else should probably review the `.iss` file. I'm not even sure what it does..

emintham:
+1 otherwise

sdetools/modules/sync_ldap/ldap_module.py
geoffwhittington: Because the logic down below implies that it will only sync based on the entries. i.e. there's no differentiation between an empty list and a null list

install/setup.iss
geoffwhittington: nope, just me! This file is ok - I've already tested it. No one code reviews this....

sdetools/modules/sync_ldap/ldap_module.py
geoffwhittington: I don't know - I don't know the code that well, I just know Python doesn't guarantee what state that variable is in

sdetools/modules/sync_ldap/ldap_module.py
geoffwhittington: OH, I'm just copying what was there .. or am I?

install/setup.iss
geoffwhittington: .iss files are Inno Setup installer files

sdetools/extlib/listen/signal_handler.py
geoffwhittington: can't make references to signal attributes if they don't make sense for the OS. Keeping `signal.SIGHUP` in the program flow will break the app, for instance.

sdetools/modules/sync_ldap/ldap_module.py
geoffwhittington: Also, I failed on the tests because I didn't account for these variables being falsy

PR #455, opened 2017-03-08
Use available certificates for LDAPS

geoffwhittington:


geoffwhittington:
@emintham I believe we're going to need this for successful LDAPS

geoffwhittington:
Have you tested against ldaps much re: certs?

emintham:
not at all.

PR #456, opened 2017-03-11
Make sure values are non-empty before updating default user schema

geoffwhittington:


JeffBain:
Mostly looks good here, just wasn't sure if we missed the odd logging statement or not. If those are intentional, +1.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -897,9 +898,12 @@ def deactivate_stale_users(self, tracked_users, sd_users):
 
         for user in users_to_deactivate:
             try:
-                self.sde_plugin.update_user(**user)
+                self.sde_plugin.update_user(id=user['id'], is_active=False, groups=user['groups'])
                 deactivated_user_count += 1
 
+                error_msg = ("SD Elements user '%s' was not found in the "
+                             "directory and is now disabled")
+                logger.debug(error_msg, user["email"])
```
geoffwhittington: self.emit() will trigger output in the celery logs. logger goes to stdout - so depending on how you run the command you get different output

geoffwhittington: There is no "debug" statement for emit() so we want to be judicious on when we use emit

geoffwhittington:
Great, thanks! You can merge and create a new release - I'm not sure how you guys are doing this anymore...

geoffwhittington:
@JeffBain you didn't click squash and merge! I just assumed my noisy commits would go quietly into that good night... :)

PR #458, opened 2017-03-17
Fix unicode issues with ldap group names

RouganStriker:
Refs #SDE-3738


harry-kim:
can we test this?

emintham:
+1

JeffBain:
Code change looks fine. Harry makes a good point about testing if possible, though I assume maybe this is tricky to test since the place we run into issues is during the sync, correct?

RouganStriker:
Yea, we can manually test it using LDAP groups with unicode characters in the anme

PR #460, opened 2017-03-18
LDAPS is LDAP plus Start TLS

geoffwhittington:


geoffwhittington:
This fixes a bug trying to connect to a server using ldaps

PR #462, opened 2017-03-20
Fix LDAPS connections and include more detail in bind errors

geoffwhittington:


RouganStriker:
+1, @emintham what do you think of the change?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -219,8 +219,10 @@ def initialize(self):
                 if k in self.user_schema and self.config['user_schema'][k]:
                     self.user_schema[k] = self.config['user_schema'][k]
 
+        # At the moment, LDAP is the only supported method
+        #   LDAPS is actually LDAP with starttls turned on
         self.connection = ldap.initialize(
-            "%s://%s" % (self.ldap_method, self.config["ldap_server"])
+            "ldap://%s" % self.config["ldap_server"]
```
emintham: If we're removing the ability for the user to control ldap_method, we should probably remove it as an option.

RouganStriker: We are technically using it as a hacky way to detect if the user wants the start tls option. If we remove the ldap_method we will need to add the start tls option

geoffwhittington: We can add the start tls option instead, I just don't know people's appetite to make that change given how long LDAP is taking

RouganStriker: It's not a lot of work to add the start_tls option. Do we need to maintain backwards compatibility, ie keep ldap_method and account for ldaps usage?

RouganStriker:
Removed ldap method and add start_tls in #463

RouganStriker:
I think we can close this PR since the changes have been addressed in #463

PR #463, opened 2017-03-21
Remove LDAP method param. Add starttls param.

RouganStriker:
Refs #SDE-3778


sdetools/modules/sync_ldap/tests/ldap_module_test.py
```diff
@@ -84,6 +84,9 @@ def result3(self, msg_id):
             [SimplePagedResultsControl(size=1, cookie=cookie)]
         )
 
+    def start_tls_s(self):
+        pass
```
RouganStriker: This is part of the mocked LDAP connection object; we need to define this or else the tests will fail because it will try to call start_tls_s on our mock object

JeffBain: Ahh fair. Maybe just add a comment to that effect so it's clear what's up.

RouganStriker: I added a short docstring to the Mock object saying that we are stubbing out methods used in syncs

PR #464, opened 2017-03-21
Fix groupless SDE users not getting deactivated

RouganStriker:
Refs #SDE-3767


sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -778,14 +778,12 @@ def update_user(self, sd_user, dir_user):
         sde_groups = self.translate_ldap_group_to_sde_group(
             dir_user["groups"], self.config['group_mapping']
         )
-        groups = [grp['id'] for grp in sde_groups]
+        new_groups = [grp['id'] for grp in sde_groups]
 
         # Deactivate user
-        deactivation = self.config["deactivation"]
-        user_active = sd_user['is_active']
-        without_groups = not sd_user['groups'] and not sde_groups
```
JeffBain: Ahh, I assume this was the problem line?

JeffBain:
+1. Looks fine to me.

sdetools/modules/sync_ldap/ldap_module.py
RouganStriker: Partly, without_groups would never be evaluated to False because sde_groups will always contain a group (otherwise there would be no LDAP groups to sync)

Gee19:
+1

funkaoshi:
+1

PR #465, opened 2017-03-22
Initialize LDAP TLS parameter for console use

geoffwhittington:


geoffwhittington:
@RouganStriker  Just a small change when sync runs from the command-line

RouganStriker:
+1

PR #466, opened 2017-03-22
Allow users to confirm deactivations by default

RouganStriker:
* Fixed default value for default_stale_users param

Refs #SDE-3789


RouganStriker:
@geoffwhittington do you have any ideas why the appveyor check failed? Is it due to the default value change?

funkaoshi:
+1

PR #467, opened 2017-03-23
WIP: Add troubleshoot flag which enables file retention

aeldamaty:
Downloads all files to a specified directory and saves http reqs and
responses.

Refs #SUP-24


sdetools/sdelib/restclient.py
```diff
@@ -4,13 +4,17 @@
...
 
 
 from commons import json, Error, UsageError, urlencode_str
 from sdetools.extlib import http_req
 from sdetools.extlib.ntlm import HTTPNtlmAuthHandler
+from sdetools.sdelib import commons
+from pprint import pformat
 
 logger = logging.getLogger(__name__)
+optional_logger = logging.getLogger('optional_logger')
```
geoffwhittington: Can we add a troubleshooting file handler for the existing logger?

sdetools/sdelib/restclient.py
```diff
@@ -375,6 +380,20 @@ def call_api(self, target, method=URLRequest.GET, args=None, call_headers={}, au
         handle.close()
         result = self.parse_response(result, dict(handle.headers))
 
+        if commons.troubleshoot_path:
+            if not self.handler:
+                self.handler = logging.FileHandler(os.path.join(commons.troubleshoot_path, './http_req_and_res.log'))
+                formatter = logging.Formatter('%(asctime)s %(levelname)s %(message)s')
+                self.handler.setFormatter(formatter)
+                optional_logger.addHandler(self.handler)
+                optional_logger.setLevel(logging.DEBUG)
```
geoffwhittington: What if we added a VERBOSE mode to the existing logger which would output the request and response?

geoffwhittington:
I think you want to remove the .fpr and .log file from the commit

sdetools/analysis_integration/base_integrator.py
```diff
@@ -56,6 +58,19 @@ def clear(self):
     def can_parse_file(self):
         return False
 
+    @staticmethod
+    def download_file_to_troubleshooting_directory(the_file):
```
geoffwhittington: Perhaps `capture_file_for_troubleshooting` ?

sdetools/analysis_integration/base_integrator.py
```diff
@@ -56,6 +58,19 @@ def clear(self):
...
+        # Check if file is of type file, grab the name, otherwise the_file is assumed to be string
+        name = the_file
+        if isinstance(the_file, file):
+            name = the_file.name
+
+        shutil.copy(name, commons.troubleshoot_path)
+
+    def check_for_troubleshoot_mode(self, file):
```
geoffwhittington: You could eliminate this method and put its contents to the top of the other method

sdetools/analysis_integration/base_integrator.py
```diff
@@ -135,6 +150,8 @@ def process_archive(self, zip_archive):
 
         logger.debug("Processing archive file: %s" % zip_archive)
 
+        self.check_for_troubleshoot_mode(zip_archive)
```
geoffwhittington: Just call the method and decide in there whether to actually do the work

sdetools/sdelib/conf_mgr.py
```diff
@@ -128,6 +128,7 @@ class Config(object):
         'args': None,
         'proxy_auth': '',
         'cert_loc': '',
+        'troubleshoot': ''
```
geoffwhittington: `troubleshoot`  could be called `session_capture_directory` ?

sdetools/sdelib/conf_mgr.py
```diff
@@ -329,6 +333,12 @@ def parse_args(self, cmd_inst):
 
             setup_media_path(self['media_path'])
 
+        if self['troubleshoot']:
+            if not os.path.isdir(self['troubleshoot']):
```
geoffwhittington: If an integration client/server goes into troubleshooting mode its going to have a TON of files in it. I agree we want to make sure that the directory exists. Our solution should make it easy to collect and find relevant troubleshooting files for a session.

geoffwhittington:
@sibartlett it would be good to have your thoughts on this

sdetools/sdelib/restclient.py
geoffwhittington: Hmm I see the problem. We only want VERBOSE for troubleshooting not the normal logging. I guess we need a separate logger in that case. What if we had a `troubleshooting_logger` that would have the NULL handler (or no handler) by default and it would always output VERBOSE. Then if troubleshooting is turned on then we attach a real file handler to the `troubleshooting_logger` - does this make any sense?

sdetools/sdelib/restclient.py
aeldamaty: I think i like having a separate logger better. I changed the name to be troubleshooting_logger which I think is better.

sibartlett:
I've updated the format of the HTTP logging:

```
2017-04-04 17:59:15,624 INFO HTTP

GET http://localhost:8000/api/v2/users/me/
    
    Host: localhost:8000
    Authorization: Token b894d938aa0fc840a0ef5ea419ee1576f1b8a319
    User-agent: Python-urllib/2.7

200 OK
    
    Server: WSGIServer/0.1 Python/2.7.13
    Content-Length: 81
    Expires: -1
    Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-eval' 'unsafe-inline'; img-src 'self' data: blob:; style-src 'self' 'unsafe-inline'
    Content-Language: en
    X-Content-Options: nosniff
    Strict-Transport-Security: max-age=31536000 ; includeSubDomains ; preload
    Vary: Accept-Language, Cookie
    Pragma: no-cache
    X-UA-Compatible: IE=Edge,chrome=1
    Allow: GET, PUT, PATCH, HEAD, OPTIONS
    Cache-control: no-cache, no-store, max-age=0, must-revalidate
    Date: Tue, 04 Apr 2017 21:59:15 GMT
    X-Frame-Options: SAMEORIGIN
    Content-Type: application/json
    Set-Cookie:  sessionid=vwof4mtk5b9km6vfrop5vow7knwglkcp; httponly; Path=/
    
    {"id":1,"email":"admin@example.com","first_name":"Admin","last_name":"Testerton"}

2017-04-04 18:11:48,565 INFO HTTP

POST http://localhost:8000/api/v2/projects/1/tasks/1-T375/analysis-notes/
    
    Content-length: 86
    Content-type: application/json
    Host: localhost:8000
    Authorization: Token b894d938aa0fc840a0ef5ea419ee1576f1b8a319
    User-agent: Python-urllib/2.7
    
    {"behaviour": "replace", "confidence": "high", "analysis_session": 27, "findings": []}

201 Created
    
    Server: WSGIServer/0.1 Python/2.7.13
    Content-Length: 249
    Expires: -1
    X-Debug-Query-Count: 48
    Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-eval' 'unsafe-inline'; img-src 'self' data: blob:; style-src 'self' 'unsafe-inline'
    Content-Language: en
    X-Content-Options: nosniff
    Strict-Transport-Security: max-age=31536000 ; includeSubDomains ; preload
    Vary: Accept-Language, Cookie
    Pragma: no-cache
    X-UA-Compatible: IE=Edge,chrome=1
    Allow: GET, POST, HEAD, OPTIONS
    Cache-control: no-cache, no-store, max-age=0, must-revalidate
    Date: Tue, 04 Apr 2017 22:11:48 GMT
    X-Frame-Options: SAMEORIGIN
    Content-Type: application/json
    Set-Cookie:  sessionid=wda1l4zo16efojj9kfyfzt8kyi9c0hre; httponly; Path=/
    
    {"id":1119,"analysis_session":27,"analysis_type":{"slug":"fortify","name":"Fortify"},"applicable":true,"automatic":true,"behaviour":"replace","confidence":"high","created":"2017-04-04T22:11:47.956592Z","findings":[],"finding_ref":"","status":"pass"}
```

sdetools/sdelib/restclient.py
```diff
@@ -9,8 +9,10 @@
 from commons import json, Error, UsageError, urlencode_str
 from sdetools.extlib import http_req
 from sdetools.extlib.ntlm import HTTPNtlmAuthHandler
+from sdetools.sdelib import session_capture
 
 logger = logging.getLogger(__name__)
+troubleshooting_logger = logging.getLogger('troubleshooting_logger')
```
geoffwhittington: I think we can drop this? I don't see this used anywhere (but I wonder how the linter didn't catch this....)

sibartlett: Good catch! The linter probably didn't catch it because it's in global scope.

sibartlett:
Files are now saved to a timestamped sub-directory on each run. The directory is in ISO date format.

geoffwhittington:
@sibartlett can you check my commits to make sure they make sense? I fixed a few issues that came out of testing on Windows and using the config file.

PR #468, opened 2017-03-23
Use group id instead of group name for group_mapping in LDAP Connections

RouganStriker:
Refs #SDE-3782


geoffwhittington:
Using the id means we break from the original idea that sdetools could be run from the command-line as a shell user. Not many people are going to the know the id for a group in SDE

Referential integrity is a tough thing to do without the id's I get it. @sibartlett you cool with this new approach?

geoffwhittington:
+1

PR #469, opened 2017-03-28
Add invalid option values check on field validate

Gee19:
Accept options values as strings with comma delimeter.

Update sde_verification_filter default to match correct values expected by the api.


lmontopo:
Nice dude, I think this looks really great! Good find.  
+1

PR #474, opened 2017-04-06
Optimize LDAP sync user creation

RouganStriker:
* Added support for multipart file upload to restclient

Refs #SDE-3896


sdetools/sdelib/sdeapi.py
```diff
@@ -92,6 +92,26 @@ def get_users(self, myself_only=False):
...
+                 user.get('first_name'),
+                 user.get('last_name'),
+                 user.get('role', ''),
+                 ';'.join(user.get('groups', []))
+            )
+            for user in users_data
+        ]))
+        # Add some attributes to this fake file
+        csv.name = "users.csv"
+        csv.content_type = "multipart/form-data; boundary=2557e666c814471fa1c007c6c3aaa579"
```
emintham: what's this magical boundary value?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -932,32 +933,39 @@ def sync_new_users_from_ldap(self, untracked_users, directory_users):
...
+            })
 
-                update_failures.append([
-                    error_message % user,
-                    user,
-                    str(dir_user),
-                ])
+        try:
+            for i in xrange(0, int(ceil(len(untracked_users)/2000.0))):
+                # The bulk user endpoint is limited to 2000 users
```
emintham: Where does this restriction come from? Is it likely to change in the future?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -932,32 +933,39 @@ def sync_new_users_from_ldap(self, untracked_users, directory_users):
...
+                start = i * 2000
+                end = start + 2000
+                print "Updating users {0}-{1}".format(start, end)
```
emintham: logger.info here maybe?

sdetools/sdelib/restclient.py
```diff
@@ -264,6 +269,8 @@ def call_api(self, target, method=URLRequest.GET, args=None, call_headers={}, au
         """
         if not self.opener:
             self.post_conf_init()
+        if not call_headers:
```
emintham: Might want to make this call_headers is not None instead. {} is also falsy.

sdetools/sdelib/sdeapi.py
```diff
@@ -52,10 +52,12 @@ def connect(self):
         return bool(self.me)
 
     def call_api(self, target, method=URLRequest.GET, args=None,
-                 call_headers={}, auth_mode=None, page_size=0):
+                 call_headers=None, auth_mode=None, page_size=0, multipart=False):
+        if not call_headers:
```
emintham: Same comment re: call_headers here.

sdetools/sdelib/sdeapi.py
```diff
@@ -77,8 +79,6 @@ def call_api(self, target, method=URLRequest.GET, args=None,
 
             page_num += 1
 
-        return results
```
emintham: Why was this removed?

setup.py
```diff
@@ -67,6 +67,7 @@ def set_py2exe_options():
     url='https://github.com/sdelements/sdetools',
     install_requires=[
         "yantra",
+        "poster",
```
emintham: I think we're trying to put all our dependencies into our vendor directory @geoffwhittington ?

emintham:
What does the endpoint do with multi-part files? Does it wait for all the parts to be present before actually doing any imports or does it do it part by part?

If we've already validated the data on the LDAP sync side and don't need to re-validate on the users endpoint side, I wonder if we might get a crude resume-able sync by sending each part individually. e.g. break users into n CSVs, send the CSVs one by one and log our progress. If network fails or w/e causes us to stop, we can resume from where we last finished (manually).

Also, we can potentially have multiple processes work on importing the users since they are working on different CSVs. Caveat here is that the CSV we generate needs to validate that each row has a unique user and that we aren't in a data race to edit the same row.

sdetools/modules/sync_ldap/ldap_module.py
RouganStriker: I added this limit in when I noticed the API timing out if you try to bulk create too many users at once. Right now 2k users take maybe 2 minutes, if this isn't too long we can probably bump  up the limit.

sdetools/sdelib/sdeapi.py
RouganStriker: This code was unreachable due to the while loop

RouganStriker:
> What does the endpoint do with multi-part files? Does it wait for all the parts to be present before actually doing any imports or does it do it part by part?
I think by the time our endpoint code runs, everything will be completely there.

> If we've already validated the data on the LDAP sync side and don't need to re-validate on the users endpoint side, I wonder if we might get a crude resume-able sync by sending each part individually. e.g. break users into n CSVs, send the CSVs one by one and log our progress. If network fails or w/e causes us to stop, we can resume from where we last finished (manually).
Technically it is already resume-able; we are performing real creates in batches of 2000 that don't get rolled back if subsequent import fails. So re-running the sync will not re-create users that were already created. And I don't think we should be validating the data on the LDAP (sdetools?) side. Especially for remote integrations where the delay between the time of check and time of use is much greater.

> Also, we can potentially have multiple processes work on importing the users since they are working on different CSVs. Caveat here is that the CSV we generate needs to validate that each row has a unique user and that we aren't in a data race to edit the same row.
Unless we are running multiple LDAP syncs at the same time the data race shouldn't be an issue since we are constructing our user dictionaries using the email as a key.

sdetools/sdelib/sdeapi.py
RouganStriker: Some smart person on the internet can probably explain this better than me but it is used to split post data for multipart/form-data content types.

setup.py
geoffwhittington: That was the dream. It's controversial to some. Check with Simon on final word

sdetools/sdelib/sdeapi.py
geoffwhittington: Should be able to assign any value 

http://stackoverflow.com/questions/3508338/what-is-the-boundary-in-multipart-form-data

sdetools/modules/sync_ldap/ldap_module.py
geoffwhittington: It 2k too much for a busy server? Do we need to make it a *gasp* an option? Would be good to make it a constant in the code anyways

RouganStriker: Not sure, it takes maybe 2mins right now on my local machine. We can probably test it out on richmond once the changes are there.

setup.py
RouganStriker: The extlib directory?

sdetools/modules/sync_ldap/ldap_module.py
geoffwhittington: I'm not sure Richmond or any server we have is a good test of busy-ness. The only person who really uses it is Vey. Maybe 2K isn't such a bad number. But I think we should still make it a constant in the code

geoffwhittington:
Is this still a WIP?

RouganStriker:
No, it is ready for review

geoffwhittington:
There seems to be a missing dependency `poster`?

RouganStriker:
I added poster dep to the setup.py file; should I be pulling a copy of it in?

RouganStriker:
@sibartlett / @emintham can I get a +1?

emintham:
+1

sibartlett:
`poster` should've been copied into `extlib` instead of being a pip dependency.

I've now fixed it.

PR #475, opened 2017-04-08
Feature/appscancloud

keyvanm:
- Created a base class for appscan cloud modules
- Created an implementation of the above for appscan dynamic website analyzer

TODO: 

- [ ] Refactor `import_appscan_mobile_analzyer` to use the new base class


geoffwhittington:
If there is any directory under `modules` it has to be an actual module - not an abstract module for the purposes of code re-use. 

My preference is to RY, i.e. lets Repeat Ourselves. The base class may be useful but we're not 100% sure what the differences are between the mobile/dynamic/static analyzers so I think its ok to repeat ourselves. The base class is a new architecture change and I don't think we're ready for that yet. There's time to refactor in the future.

I would literally copy/paste the mobile analyzer code into a distinct dynamic analyzer. Rename it for dynamic, make sure the XML parser picks up the right stuff and call it a day.

geoffwhittington:
@sibartlett what are your thoughts ? I'll defer to whats on your mind.

PR #476, opened 2017-04-10
Raise friendly error message when issue type can’t be retrieved for a…

sibartlett:
… project

Refs #SUP-30


sdetools/modules/sync_jira/jira_rest.py
```diff
@@ -93,6 +93,10 @@ def setup_fields(self, issue_type_id, new_issue_template):
             raise AlmException('Could not retrieve fields for JIRA project: %s. %s' %
                                (self.config['alm_project'], error))
 
+        if not meta_info['projects']:
+            raise AlmException('Do not have permission to access issue type "%s" for JIRA project: %s.' %
```
geoffwhittington: `You do not have ....` or `Permission error accessing issue type ....` ?

geoffwhittington: https://developer.atlassian.com/static/rest/jira/6.1.html#d2e1328

"....Projects will not be returned if the user does not have permission to create issues in that project." << That's what we're seeing

sibartlett: Updated error message.

geoffwhittington:
+1

PR #478, opened 2017-04-13
Added support for IBM Application Security on Cloud

sibartlett:
Refs #SDE-3882


sdetools/modules/import_asoc/asoc_integrator.py
```diff
@@ -1,16 +1,20 @@
...
+__all__ = ['AsocIntegrator']
 
 
-class AppScanMobileAnalyzerIntegrator(BaseIntegrator):
-    TOOL_NAME = "appscan_mobile_analyzer"
-    DEFAULT_MAPPING_FILE = 'sde_appscan_mobile_analyzer_map.csv'
+class AsocIntegrator(BaseIntegrator):
+    TOOL_NAME = 'IBM Application Security on Cloud'
+    MAPPING_FILES = {
+        'dynamic_analyzer': 'sde_appscan_mobile_analyzer_map.csv',
```
geoffwhittington: These .csv file names should reflect the tools, I think - i.e.

- sde_appscan_dynamic_analyzer_map.csv
- sde_appscan_mobile_analyzer_map.csv
- sde_appscan_static_analyzer_map.csv

sdetools/modules/import_asoc/asoc_xml_importer.py
```diff
@@ -57,14 +61,16 @@ def endElement(self, name):
...
 
 
-class AppScanMobileAnalyzerXMLImporter(BaseXMLImporter):
+class AsocXMLImporter(BaseXMLImporter):
 
     def __init__(self):
-        super(AppScanMobileAnalyzerXMLImporter, self).__init__()
-        self.edition = 'mobile_analyzer'
+        super(AsocXMLImporter, self).__init__()
+        self.technolgies = {
```
geoffwhittington: Seems like this should work? This is a break from what we normally do: assume nothing about the technology covered by a scanner and instead just consider the weaknesses it identifies.

I don't know much about the schema to say whether this may change at all... Perhaps we can ask IBM?

sibartlett: The XML file outputted by ASOC contains a `technology` attribute. It indicates which analyzer was used. DAST for dynamic, SAST for static, and Android or iOS for mobile.

I figured that it was the most reliable attribute to use, in order to determine the analyzer used.

sibartlett:
Opened a new pull request which contains both ASoC and AppScan Source. #485

PR #479, opened 2017-04-17
Added support for IBM AppScan Source

sibartlett:
Refs #SDE-3883


sibartlett:
Opened a new pull request which contains both ASoC and AppScan Source. #485

PR #480, opened 2017-04-24
Feature/sdetools refactor

geoffwhittington:


sdetools/alm_integration/github_integration_plugin.py
```diff
@@ -0,0 +1,98 @@
+from alm_integration_plugin_base import (BaseALMIntegration,
+                                         FieldType, Section, IntegrationField)
+
+from sdetools.modules.i18n import _
+
+
+class GithubIntegrationPlugin(BaseALMIntegration):
```
geoffwhittington: @Gee19 @lmontopo @sibartlett Any reason why these don't live inside their respective plugin directories?

sdetools/alm_integration/jira_integration_plugin.py
```diff
@@ -0,0 +1,176 @@
...
+            label=_('Jira close transition'),
+            field_type=FieldType.STRING,
+            cli_description='Close transition in JIRA',
+            placeholder='Close Issue'
+        )
+        # NOTE: the ones below are common to several plugins
+        # when other plugins are added that use them, these should be moved
+        # to the base and then just put in sections of the plugins that have
+        # them.
+        self.title_format = IntegrationField(
```
geoffwhittington: Will these move to the base before merge?

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,543 @@
...
+            cli_description='Show progress',
+            default='False'
+        )
+        self.test_alm = IntegrationField(
+            name='test_alm',
+            field_type=FieldType.STRING,
+            cli_description='Test Alm "server", "project" or "settings" '
+                            'configuration only'
+        )
+        self.alm_custom_fields = IntegrationField(
```
geoffwhittington: This is not a CLI-only field

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,543 @@
...
+        ]
+
+    def get_cli_only_fields(self):
+        return [
+            self.start_fresh,
+            self.alm_standard_workflow,
+            self.selected_tasks,
+            self.show_progress,
+            self.test_alm,
+            self.alm_custom_fields,
```
geoffwhittington: This is an important field that should live in the UI

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -0,0 +1,543 @@
...
+        )
+        self.howtos_in_scope = IntegrationField(
+            name='how_tos_in_scope',
+            label=_('Include code sample How-To\'s in task descriptions'),
+            field_type=FieldType.BOOLEAN,
+            cli_description='Whether or not HowTos should be included',
+            default=False
+        )
+
+        # TODO: Should we be hardcoding phase?
```
geoffwhittington: misplaced comment?

sdetools/alm_integration/jira_integration_plugin.py
```diff
@@ -0,0 +1,176 @@
...
+            field_type=FieldType.STRING,
+            cli_description='Statuses that signify a task is Done in JIRA',
+            placeholder='Resolved,Closed'
+        )
+        self.jira_issue_type = IntegrationField(
+            name='jira_issue_type',
+            label=_('Issue Type:'),
+            field_type=FieldType.STRING,
+            cli_description='IDs for issues raised in JIRA',
+            placeholder='Bug'
```
geoffwhittington: Does Placeholder ==  Default? The underlying command defines the default as "Bug".

sdetools/alm_integration/github_integration_plugin.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @property
+    def id(self):
+        return 'github'
+
+    @property
+    def name(self):
+        return 'Github'
```
geoffwhittington: GitHub

sdetools/alm_integration/jira_integration_plugin.py
lmontopo: no placeholder does not equal default.  I can't think of a usecase off the top of my head but I know that currently in the UI we have some cases where there is a placeholder and NO default.  I think this is the case for the server uri actually.

sdetools/alm_integration/jira_integration_plugin.py
lmontopo: Anything that is common among more than one integration will be moved to the baseclass.

sdetools/alm_integration/alm_integration_plugin_base.py
lmontopo: yes it is misplaced.  This should be above the alm_phases, I believe. Thanks for catching this.

sibartlett:
Opened new pull request: #539

PR #482, opened 2017-04-24
moved alm_custom_fields to a first tier field

geoffwhittington:


sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -182,79 +183,151 @@ def __init__(self):
...
         )
         self.server = IntegrationField(
             name='server',
             label=_('Server'),
             required=True,
             field_type=FieldType.STRING,
             placeholder='bugs.server.com',
-            cli_description='the url of the server'
+            cli_description='the url of the server',
+            help_text=_('The domain name or IP address of the server')
```
lmontopo: @geoff: just wondering what you had in mind when you added this `help_text`.  Is your thought that it should appear in the UI or in the command line?  Also, I sort of think that the text you've written is more of a 'description' rather than 'help text'.  However, putting it as a description (or even help text for that matter) would indicate that we want it to appear in the UI.  Do we want this?

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -182,79 +183,151 @@ def __init__(self):
...
             placeholder='apitoken',
+            help_text=_('Token used to authenticate to %s' % self.name)
         )
         self.conflict_policy = IntegrationField(
             name='conflict_policy',
-            label=_('Choose Authoritative Source'),
+            label=_('Authoritative Source'),
             field_type=FieldType.RADIO,
-            cli_description='Conflict policy to use',
+            cli_description='Authoritative source',
```
lmontopo: We could have made a mistake, but just FYI @jordan and I made an effort to make the `cli_description` whatever it is currently.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -182,79 +183,151 @@ def __init__(self):
...
             default='alm',
             options=[
-                Option('ALM', 'alm'),
+                Option(self.name, 'alm'),
                 Option('SD Elements', 'sde'),
                 Option('Last Status Change', 'timestamp', disabled=True)
             ],
+            help_text=_('The de facto source for the status of an SD Elements task')
```
lmontopo: is this help_text or description? it can be tricky to differentiate them sometimes, but to me the above sounds more like description.

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -450,7 +537,8 @@ def get_task_sync_section(self):
                 self.alm_phases,
                 self.sde_tags_filter,
                 self.sde_verification_filter
-            ])
+            ],
+            description='Select which SD Elements tasks to synchronize.')
```
lmontopo: In order for these descriptions to appear anywhere we'll need to update the frontend to expect them.  Is this something we should do (maybe create a ticket for it in this sprint or the next) @sibartlett ?

PR #485, opened 2017-04-26
ASoC and AppScan Source

sibartlett:
Added support for:

* Application Security on Cloud - #SDE-3882
* AppScan Source - #SDE-3883


sdetools/modules/import_asoc/asoc_xml_importer.py
```diff
@@ -1,14 +1,17 @@
 from sdetools.analysis_integration.base_integrator import BaseXMLImporter, BaseContentHandler
 
 
-class AppScanMobileAnalyzerXMLContent(BaseContentHandler):
-    def __init__(self):
+class AsocXMLContent(BaseContentHandler):
+    def __init__(self, technolgies):
+        self.technolgies = technolgies
```
RouganStriker: Is this a typo? `technologies`

sdetools/modules/import_asoc/asoc_integrator.py
```diff
@@ -51,6 +56,9 @@ def parse_report_file(self, report, report_type):
 
         importer.parse(report)
 
+        self.TOOL_NAME = importer.technology
+        self.DEFAULT_MAPPING_FILE = self.MAPPING_FILES[self.TOOL_NAME]
```
RouganStriker: I think we should do `self.MAPPING_FILES.get(self.TOOL_NAME)` in case `importer.technology` is `None`

sibartlett: I think the better solution would be to raise an exception if importer.technology is `None`.

sdetools/modules/import_asoc/asoc_xml_importer.py
```diff
@@ -57,14 +59,15 @@ def endElement(self, name):
...
-        super(AppScanMobileAnalyzerXMLImporter, self).__init__()
-        self.edition = 'mobile_analyzer'
+        super(AsocXMLImporter, self).__init__()
+        self.technology = None
 
     def _get_content_handler(self):
-        return AppScanMobileAnalyzerXMLContent()
+        return AsocXMLContent()
 
-    def get_edition(self):
```
RouganStriker: This was used by the appscan integrator; are these importers not being used there?

geoffwhittington: No, not here -  I think it was a copy/paste error from earlier

geoffwhittington: This is a net new sdetools module unaffiliated to "import_appscan" - it was here originally incorrectly

geoffwhittington:
+1

PR #486, opened 2017-04-26
Add all plugins as plugin-metadata.py

Gee19:


sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -182,78 +183,150 @@ def __init__(self):
...
             label=_('{0} Username'.format(self.name)),
             placeholder='username',
             required=True,
-            field_type=FieldType.STRING
+            field_type=FieldType.STRING,
+            help_text=_('Username authorized to connect with the server')
         )
         self.password = IntegrationField(
             name='alm_pass',
             label=_('{0} Password'.format(self.name)),
```
geoffwhittington: Do we need to qualify the tool? If so, why not qualify all the fields in the same way?

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -415,20 +496,23 @@ def get_connection_details_section(self):
...
+                       fields=[self.username, self.password],
+                       description=_('Enter the credentials needed to authenticate to the server.'))
 
     def create_token_credentials_section(self):
         return Section(id='credentials',
                        label=_('Credentials'),
                        optional=False,
-                       fields=[self.token])
+                       fields=[self.token],
+                       description=_('Enter the token needed to authenticate to the server.'))
```
geoffwhittington: `description=_('Token needed to authenticate to the server.'))`

sdetools/alm_integration/alm_integration_plugin_base.py
```diff
@@ -415,20 +496,23 @@ def get_connection_details_section(self):
...
+            description=_('Enter the connection details for the server.')
         )
 
     def create_username_password_credentials_section(self):
         return Section(id='credentials',
                        label=_('Credentials'),
                        optional=False,
-                       fields=[self.username, self.password])
+                       fields=[self.username, self.password],
+                       description=_('Enter the credentials needed to authenticate to the server.'))
```
geoffwhittington: description=_('Credentials needed to authenticate to the server.'))

sdetools/alm_integration/alm_integration_plugin_service.py
```diff
@@ -17,11 +18,14 @@ def get_plugins(cls, plugins_path=None):
         plugins = []
 
         for f in os.listdir(path):
-            full_path = os.path.join(path, f)
+            full_path = os.path.join(os.path.abspath(path), f)
             plugin_path = os.path.join(full_path, 'plugin-metadata.py')
             if os.path.isdir(full_path) and os.path.isfile(plugin_path):
                 tmp = imp.find_module("plugin-metadata", [full_path])
                 try:
+                    full_path = full_path.replace(sdelib.commons.base_path, 'sdetools')
```
geoffwhittington: Interesting, I expected you'd change the `INTEGRATION_PLUGIN_PATH` value above to use `sdelib.commons.base_path`, i.e.

```
INTEGRATION_PLUGIN_PATH = os.path.join(sdelib.commons.base_path, 'sdetools/modules')
```
and the rest would fall into place? Perhaps that would make a difference?

sdetools/alm_integration/alm_integration_plugin_base.py
Gee19: Looks like we are only doing it for the Username/Password/Token fields. Should I remove them?

geoffwhittington: My vote is yes, seems redundant

sibartlett: We have a separate ticket to go through and review all fields (including labels etc), for each individual integration. It'll probably be one of the last things we do before it all gets merged.

sdetools/modules/sync_github/plugin-metadata.py
```diff
@@ -1,10 +1,10 @@
-from sdetools.alm_integration.alm_integration_plugin_base import (BaseALMIntegration, FieldType,
-                                                                  Section, IntegrationField)
+from sdetools.alm_integration.alm_integration_plugin_base import (BaseALMIntegration,
+                                                                  FieldType, Section, IntegrationField)
 
 from sdetools.modules.i18n import _
 
 
-class GithubIntegrationPlugin(BaseALMIntegration):
+class PluginMetadata(BaseALMIntegration):
```
lmontopo: I don't like how all of these classes are named the exact same for each alm.  Its confusing.  I think we should do something like `GithubMetadata` if we want to use the name `metadata` instead of `plugin` (which is a good call since we already have a `plugin`).

sdetools/modules/sync_github/plugin-metadata.py
```diff
@@ -19,13 +19,15 @@ def command(self):
         return 'sync_github'
 
     def __init__(self):
-        super(GithubIntegrationPlugin, self).__init__()
+        super(PluginMetadata, self).__init__()
 
         # update alm project field
-        self.alm_project.help_text = _('GitHub Help Text Here')
+        self.alm_project.help_text = _('Repository name')
+        self.alm_project.description = self.alm_project.help_text
```
lmontopo: do we need a description here?  Especially if its the exact same text as the help text... seems like we should just not include description.

sdetools/modules/sync_hp_alm/plugin-metadata.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @property
+    def command(self):
+        return 'sync_hp_alm'
+
+    def __init__(self):
+        super(PluginMetadata, self).__init__()
+
+        # update alm project field
+        self.alm_project.help_text = _('The name of the project')
```
lmontopo: why are we updating this help text?  It already has help text which says the exact same thing but in different words. :S

sdetools/modules/sync_jira/plugin-metadata.py
```diff
@@ -0,0 +1,104 @@
...
+    @property
+    def command(self):
+        return 'sync_jira'
+
+    def __init__(self):
+        super(PluginMetadata, self).__init__()
+
+        #  update alm_project field (new method)
+        self.alm_project.label = _('JIRA Key')
+        self.alm_project.help_text = _('JIRA Help Text Here')
```
lmontopo: I had previously only put this in here because I honestly don't really understand Jira enough to know what the project is and how to find it.  If you know @geoffwhittington , feel free to let me know and I can update this now.  If not no big deal, we can come back to it.

sdetools/modules/sync_jira/plugin-metadata.py
```diff
@@ -0,0 +1,104 @@
...
+            label=_('Jira version'),
+            field_type=FieldType.RADIO,
+            default='5',
+            cli_description='Version of JIRA [e.g. 4.3.3, 5, or 6.0]',
+            options=[Option('4', '4'),
+                     Option('5+', '5')]
+        )
+        self.jira_done_statuses = IntegrationField(
+            name='jira_done_statuses',
+            label=_('States which map to DONE in SD Elements:'),
```
lmontopo: I think the colon should be removed here. I can do that.

lmontopo: Or maybe it shouldn't I am not certain actually.

sdetools/modules/sync_hp_alm/plugin-metadata.py
```diff
@@ -0,0 +1,116 @@
...
+            default='Passed'
+        )
+        self.hp_alm_new_status = IntegrationField(
+            name='hp_alm_new_status',
+            label=_('Open a story in %s with the following status' % self.name),
+            field_type=FieldType.STRING,
+            cli_description='Open a task in %s in the following state' % self.name,
+            placeholder='Not Completed',
+            default='Not Completed',
+        )
```
lmontopo: some of these things like `new_status` and `done_status` are pretty much repeats among different alm's its just that their prefixes differ.  I think that if only the prefixes of the names differ, then perhaps we should be extracting this out into the base class.  If there's a lot that changes (which is the case for some of them) then its probably not worth extracting, but i think i'll go ahead and extract out the `new_status` and `done_status` fields (unless anyone objects).

sdetools/alm_integration/alm_integration_plugin_service.py
```diff
@@ -8,8 +11,26 @@ class ALMIntegrationPluginService(object):
...
+            plugin_path = os.path.join(full_path, 'plugin-metadata.py')
+            if os.path.isdir(full_path) and os.path.isfile(plugin_path):
+
+                f, filename, description = imp.find_module("plugin-metadata", [full_path])
+
+                try:
+                    import_path = full_path.replace(sdelib.commons.base_path, 'sdetools').replace('/', '.') + '.plugin-metadata'
+
+                    plugin_module = imp.load_module(import_path, f, filename, description)
+                    plugin_class = getattr(plugin_module, "PluginMetadata")
```
lmontopo: I do think its perhaps worth while using yantra here because its already set up to do some of this magic.  I mean, I could put more work into this part so that all of the class names don't have to be the same among the different alm's, but it seems like if i'm going to put work into making things load dynamically i might as well make it work with yantra, no?  Or else if we dont' want to put that work in, I could just revert this back to the way it used to be (hardcoded) and then that would allow me to change the names so that each alm has a unique class name.  Thoughts?

lmontopo: Decision is to go back to the hardcoding way.  In the future we can use Yantra to make things load dynamically.

sdetools/modules/sync_github/plugin-metadata.py
lmontopo: i'm making this change.

sdetools/modules/sync_github/plugin-metadata.py
lmontopo: can decide this when we review validity of these plugins later.

sdetools/modules/sync_hp_alm/plugin-metadata.py
lmontopo: decided against this change.  Once i started making the change i saw that it was actually a fair bit cleaner the way it is.

sdetools/modules/sync_versionone/plugin-metadata.py
```diff
@@ -0,0 +1,81 @@
...
+        self.token.help_text = _('API Token generated within the user profile page.')
+
+        self.v1_story_type = IntegrationField(
+            name='v1_issue_type',
+            label=_('Issue Type'),
+            field_type=FieldType.STRING,
+            cli_description='Story type in %s' % self.name,
+            default='Story'
+        )
+        self.v1_done_statuses = IntegrationField(
```
lmontopo: I think i might change this to 'versionone_done_statuses' etc. Cause otherwise its a bit confusing, i feel. It deviates from the way we do other variables of this type, and it also makes me think that its actually the done status of the first version.  Thoughts?

PR #487, opened 2017-05-01
Rally auto tests

wisznie:
I was mistaken, this is a v1.39 compatible test, the upgrade for v2.0 once upgraded though is to just remove the ".js"  from the few api calls.


funkaoshi:
We gotta review this stuff!

PR #489, opened 2017-05-04
Fix LDAP sync issues

RouganStriker:
* Remove non-existant emit.debug
* Fix LDAP sync not failing on error
* Improve rendering of LDAP sync error messages

Refs #SDE-3896


funkaoshi:
Seems fine +1

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -967,14 +973,26 @@ def sync_new_users_from_ldap(self, untracked_users, directory_users):
...
+            if isinstance(e.message, dict):
+                error_message = "\n".join([
+                    "{0}: {1}".format(
+                        csv_data[start + int(row_id)]['email'],
+                        " ".join([", ".join(error) for error in errors.values()])
+                    )
+                    for row_id, errors in e.message['errors'].items()
+                ])
+
+            update_failures.append("Error creating all users: {0}".format(error_message))
```
emintham: "Error creating the following users" maybe? "all users" sounds a bit weird.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -967,14 +973,26 @@ def sync_new_users_from_ldap(self, untracked_users, directory_users):
...
         except APIError as e:
-            update_failures.append(str(e))
-            self.emit.debug(e)
-            self.emit.error("Failed to create all users")
+            # Convert the error from a dict to a nicer string
+            error_message = str(e)
+            if isinstance(e.message, dict):
+                error_message = "\n".join([
+                    "{0}: {1}".format(
+                        csv_data[start + int(row_id)]['email'],
```
emintham: Is it possible to make the row_id an int key on the dict? then we won't need to do the conversion.

emintham:
Two minor comments. +1 otherwise

sdetools/modules/sync_ldap/ldap_module.py
RouganStriker: Good point, fixed

sdetools/modules/sync_ldap/ldap_module.py
RouganStriker: I believe we are using an integer key when loading items into the dict (we get the row_id from enumerate()). Something must be stringifying it somewhere...

RouganStriker: I'm going to just leave it as-is for now

PR #492, opened 2017-05-09
Fix SSL Proxy (develop)

Gee19:
Refs #SUP-45


funkaoshi:
Should just squash the two commits down. Seems fine.

sdetools/extlib/http_req.py
```diff
@@ -209,7 +209,12 @@ def connect(self):
                                           ca_certs=self.ca_certs)
         if self.cert_reqs & ssl.CERT_REQUIRED:
             cert = self.sock.getpeercert()
-            hostname = self.host.split(':', 0)[0]
+
+            if self._tunnel_host:
```
shawnburant: if you move this logic to a method ```self._get_hostname()``` then you can create a unit test on just this functionality, and mock it out when you test the outer method.

funkaoshi:
@shawnburant's suggestion was a good one. We don't do any unit testing right now, the tests here are all big integration tests. Still good to take that approach.

funkaoshi:
Let's get this merged today.

shawnburant:
Much cleaner, 👍

PR #493, opened 2017-05-09
Fix SSL Proxy

Gee19:
Refs #SUP-45


sdetools/extlib/http_req.py
```diff
@@ -209,7 +209,12 @@ def connect(self):
                                           ca_certs=self.ca_certs)
         if self.cert_reqs & ssl.CERT_REQUIRED:
             cert = self.sock.getpeercert()
-            hostname = self.host.split(':', 0)[0]
+
+            if self._tunnel_host:
```
funkaoshi: Probably nicer to write this as:

```
host = self.host if not self._tunnel_host else self._tunnel_host
hostname = host(':', 0)[0]
```

Gee19: Fixed, thanks ram

funkaoshi:
+1

I like Shawn's suggestion in the other MR as well. Can maybe discuss that change more in the develop MR. We should merge this and make a new sdetools they can test with.

funkaoshi:
@Gee19 Oh, you'll need to update the sdetools version here as well. grep for VERSION and you should find where.

funkaoshi:
+1 we can make a new build.

PR #494, opened 2017-05-09
Clean up setup.py to not include extraneous files during build.

RouganStriker:
Refs #SDE-4109


RouganStriker:
I created a package via ./setup.py sdist and it installed without issues.

funkaoshi:
sdist does the right thing by default?

RouganStriker:
What do you mean by the right thing?

geoffwhittington:
@RouganStriker the other impact to changing this area of sdetools is that `py2exe` may not create a working Windows Remote Integration Client.

setup.py
```diff
@@ -72,6 +63,5 @@ def set_py2exe_options():
     dependency_links=[
         "git+https://git@github.com/sdelements/yantra.git@master#egg=yantra"
     ],
-    data_files=static_files,
```
sibartlett: Do we not need `data_files` anymore?

RouganStriker: I believe they are already included via the other configs

The CI tests are passing, that should mean the integration clients have the files it needs to compile too right?

sibartlett: Changes to `setup.py` won't affect the CI job. `setup.py` is only run in the context of `pip`, no?

RouganStriker: I see...I think the only code path where it could break is via the py2exe. If someone with a windows machine can verify that setup.py still works then I think the data_files are no longer needed. @geoffwhittington...

funkaoshi:
@RouganStriker can you double check the produced file, and then bump the version, and merge. We can update sde as well. (Then we'll have Jordan's change as well.)

RouganStriker:
Produced tar looks fine; did a diff compare - all the files are in the new package minus the extraneous files.

PR #495, opened 2017-05-09
Fix error message handler erroring on lists

RouganStriker:
* Fixed bulk user creation log counter always increasing by 2000

Refs #SDE-3896


funkaoshi:
+1

RouganStriker:
Also fixed a bug where the existing users check was case sensitive.

Will merge it in if those changes are good.

PR #496, opened 2017-05-10
Fix SDETools missing packages during install

RouganStriker:


funkaoshi:
Who am I to doubt the wisdom of @RouganStriker.

PR #497, opened 2017-05-15
sync_bugzilla

keyvanm:


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -880,7 +880,7 @@ def _get_matching_reference(self, sde_task):
 
         if task_references:
             task_reference = next((ref for ref in task_references
-                                   if ref['alm_connection'] == self.config['alm_reference_context']), None)
+                                   if str(ref['alm_connection']) == str(self.config['alm_reference_context'])), None)
```
keyvanm: `ref['alm_connection']` is type `int` while `self.config['alm_reference_context']` is type `str`. So the equal clause never fires and it makes the method `_get_matching_reference` return `None` and tricks the program to think the SDE task isn't linked to any alm tasks.

.gitignore
```diff
@@ -47,3 +47,5 @@ info.log
 
 #custom cert bundle
 my_root_certs.crt
+
+conf/
```
keyvanm: I put all my config files in this folder

sdetools/modules/sync_bugzilla/bugzilla_plugin.py
```diff
@@ -0,0 +1,217 @@
...
+    default_issue_template = BUGZILLA_DEFAULT_ISSUE_TEMPLATE
+    default_priority_map = BUGZILLA_DEFAULT_PRIORITY_MAP
+
+    supports_references = True
+
+    def __init__(self, config, alm_plugin):
+        """ Initializes connection to Bugzilla """
+        super(BugzillaConnector, self).__init__(config, alm_plugin)
+
+        config.opts.add('bugzilla_product', 'Product to file the bug against in Bugzilla'),
```
geoffwhittington: what do you think of using `bz_` for the prefix? Less characters to type

sdetools/modules/sync_bugzilla/bugzilla_plugin.py
```diff
@@ -0,0 +1,217 @@
...
+        create_args = self.get_new_issue(task, defaults=default_args)
+        try:
+            bugzilla_bug = self.alm_plugin.call_api("bug", method=URLRequest.POST, args=create_args)
+        except APIError as err:
+            raise AlmException('Encountered an error creating a bug: %s' % err)
+
+        return self.get_bugzilla_bug_by_id(task, bugzilla_bug["id"])
+
+    def get_note_message(self, alm_task):
+        return "Product: {product}, Bug id: {id}, Bug summary:{summary}, link: {link}".format(
```
geoffwhittington: We don't need the summary in the note, do we do this in other syncs? I didn't think we did....

keyvanm: Summary is bugzilla is the bug title

sdetools/modules/sync_bugzilla/bugzilla_plugin.py
keyvanm: Good idea!

PR #500, opened 2017-06-02
Use custom handler to allow bypass ssl in Fortify SSC

Gee19:
Refs #SUP-52


sdetools/modules/import_fortify/fortify_integrator.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, config):
         config.opts.add('analysis_user', 'Fortify SSC user', default='')
         config.opts.add('analysis_pass', 'Fortify SSC password', default='')
         config.opts.add('analysis_api_token', 'Fortify SSC authtoken (AnalysisDownloadToken permission)', default='')
+        config.opts.add('analysis_validate_cert', 'Fortify SSC bypass server certificate validation')
```
RouganStriker: I think we should default this to true

sdetools/modules/import_fortify/fortify_ssc_importer.py
```diff
@@ -31,6 +32,15 @@ def marshalled(self, context):
         header.set('axis2ns1:token', self.auth_token)
 
 
+class IgnoreSSLTransport(HttpAuthenticated):
```
RouganStriker: Can we add a small commenting explaining how this works?

RouganStriker:
I saw it work so +1 after the small changes

geoffwhittington:
thanks @RouganStriker for reviewing

Good job @Gee19 !

PR #504, opened 2017-06-09
Enable more verbose logging for LDAP sync

RouganStriker:
Refs #SUP-60


geoffwhittington:
Is this a permanent change we want to make? If so, lets merge this if possible

RouganStriker:
We added something similar to the main app to assist with debugging LDAP sso so I figure we could use this here as well. I'm not sure what is the best way to enable this from the main app though, it does expose sensitive information so it is not something we would want to enable by default.

PR #505, opened 2017-06-13
Validate Iteration Path on start-up

geoffwhittington:


sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -167,6 +167,31 @@ def alm_validate_configurations(self):
         if invalid_fields:
             raise AlmException('Invalid field(s) specified: {0}'.format(', '.join(invalid_fields)))
 
+        iteration_path = self.config['tfs_iterationpath']
+        area_path = self.config['alm_project']
```
geoffwhittington: gasp, this isn't actually the area path - but we are assuming it is.

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -167,6 +167,31 @@ def alm_validate_configurations(self):
...
+
+        iterations_urlquote_str = []
+        for path in iterations:
+            iterations_urlquote_str.append(urlquote_str(path))
+
+        self._validate_classification_node('iteration', '/'.join(iterations_urlquote_str))
+
+        self.config['tfs_iterationpath'] = '%s\\%s' % (area_path, '\\'.join(iterations))
+
+    def _validate_classification_node(self, node_type, node_path):
```
geoffwhittington: this is a function since we can use the same to validate the area path, I hope...

sdetools/modules/sync_tfs/tfs_plugin.py
```diff
@@ -167,6 +167,31 @@ def alm_validate_configurations(self):
...
+
+        if node_type not in ['area', 'iteration']:
```
geoffwhittington: could drop "area" if we're not using it, but the plan is to use it at some point

funkaoshi:
@geoffwhittington who should look at this? @shawngustaw?

geoffwhittington:
I would wait to pull this into a proper sprint - its tracked as SDE-3406.

sdetools/modules/sync_tfs/tfs_plugin.py
geoffwhittington: Since I wrote this, we have added areapath support to TFS sync - will need to change this.

geoffwhittington:
I've assigned this to  @aeldamaty in support project. Its quiet in support and I don't think we'll get to it in a sprint anytime soon.

bellkelly:
This PR looks good to me. I just had one question, but otherwise +1

bellkelly:
Tested this is in the app using QA's TFS2015 credentials. 

When the iteration path is correct, the sync progresses as it should. When the area path or iteration path is incorrect, the job fails fast and a red error icon is shown with a message like "Last Sync Errors. Correct the following errors: Invalid [Area/Iteration] Path: <the invalid path>".

Also tested using a tfs.conf file and running the sync on the command line. This is also showing the appropriate errors when an incorrect path is used, and working properly when a correct path is used.

geoffwhittington:
By the way, you may notice that the mock test is lacking. I opted for a superficial implementation: technically we would want to validate that the checks actually work based on mocked API responses, etc, etc. I'll leave it up to you guys to decide if this is the sort of test you want to add at the moment.

PR #506, opened 2017-06-13
Rename 'editable_on_project' to 'default_scope'

lmontopo:
This name is more flexible and could allow for more scopes in the
future than just two.

Refs: #SDE-4160


sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -50,7 +50,7 @@ class IntegrationField(object):
     def __init__(self, name, label='', description='', cli_description='',
                  options=None, help_text="", max_length=None,
                  field_type=FieldType.STRING, readonly=False, default=None,
-                 required=False, editable_on_project=False, placeholder=None,
+                 required=False, default_scope=None, placeholder=None,
```
Gee19: Why not default_scope=''?

lmontopo: good question! i asked myself this too, and i couldn't come up with any reason why one way is better than the other, so then i just picked one at random.  Thoughts?

Gee19:
Just one question then +1

sdetools/alm_integration/alm_metadata_plugin_base.py
Gee19: It's a string so I think default = empty string makes more sense to me, I could be wrong but I think you would only use None if it has a specific meaning to our app.

lmontopo: You might be right, but i do notice that we're not being super consistent here.  For instance, placeholder and default both default to None, and they are strings.  Perhpaps we have done those two wrong, though.  I'll wait till @sibartlett chimes in and then change what needs to be changed :)  Thanks Jordan!

sibartlett: I think I prefer `None`.. An empty string kinda has no meaning.

Gee19: Sounds good, we should modify the rest to be consistent

lmontopo: OK cool, i'll leave it as is then. Thanks!

lmontopo: @Gee19 sorry i saw this comment after i already merged.  We can discuss this and always make changes to the rest later, however I think i actually like the way we have the rest.  The ones that have emptry strings sort of make sense to me because they are like display strings.  As i said though, we can talk more in person :)

PR #511, opened 2017-06-22
Fix markdown in dynamic macros

Gee19:
Append dynamic macros in description field instead of help_text

Refs #SDE-4253


sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -168,9 +168,9 @@ class BaseALMMetadata(object):
 
     DYNAMIC_MACROS = _("Field values can use the following dynamic macros. Their "
                        "values change depending on the project being synced.\n\n"
-                       "${application} - Application name\n"
```
geoffwhittington: I suggest we shy away from including specific macros in this description since there are many macros that are possible: https://support.sdelements.com/hc/en-us/articles/204925749-JIRA-Extended-Field-Support

RouganStriker:
Typo btw: https://github.com/sdelements/sdetools/blob/b4802e608f4f368c827b6573dd8a3ffbc543bc71/sdetools/alm_integration/alm_metadata_plugin_base.py#L399

Gee19:
Fixed the typo in this PR, good catch @RouganStriker

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -168,9 +168,9 @@ class BaseALMMetadata(object):
...
-                       "${project} - Project name\n"
-                       "${context} - Value of Alm Context in an integration setup\n")
+                       "* ${application} - Application name\n"
+                       "* ${project} - Project name\n"
+                       "* ${context} - Value of Alm Context in an integration setup")
```
lmontopo: does this work?  In some cases (well, reporting) when i've used markdown, I've had to use python's "dedent" library to help because otherwise the normal python indentations mess up the markdown.

Gee19: Yep it works, we are using react-markdown in the frontend

lmontopo: cool, this looks good to me then!

lmontopo:
+1

PR #513, opened 2017-06-22
Remove cli_description from metadata

lmontopo:
Refs: #SDE-3630


Gee19:
+1, this will also get reviewed in the final verification so I think it's cool to merge

PR #514, opened 2017-06-23
Reduce memory usage of LDAP sync

RouganStriker:
* Improve Group query filter to reduce cached results
* Optimize storage of LDAP user-group data

Refs #SDE-4140


sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -337,15 +336,18 @@ def search(self, query, params, base=None, scope=None):
...
-
             except ldap.LDAPError, err:
                 error_message = ("LDAP query has raised an error.\n"
                                  "Reason: %s")
                 logger.error(error_message, err)
 
                 raise LDAPConnectorError(error_message % (str(err)))
 
-            results.extend(rdata)
+            if resultFilter:
```
RouganStriker: Just filtering out irrelevant results before cache it drastically reduced memory usage for group DNs that contain a lot of objects.

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -377,7 +377,9 @@ def get_user_params(self):
 
         return params
 
-    def get_groups(self):
+    def get_groups(self, filter_groups=None):
+        from ldap.filter import escape_filter_chars
```
JeffBain: What's the reason this is here instead of at the top of the file?

sdetools/modules/sync_ldap/ldap_module.py
```diff
@@ -454,37 +476,86 @@ def get_group_from_ldap_tuple(self, ldap_tuple):
 
         return (group_name, members, group_bind_dn)
 
-    def parse_groups(self, ldap_groups, mapping):
+    def chunk(self, iterable, chunk_size):
```
JeffBain: This feels like it might be ideal as an generator, where you can just yield the chunks instead of building the accumulated chunked list. Since it's only ever iterated over once there's no need to hold onto slices we've already processed or generate slices before we're on to that chunk.

sdetools/modules/sync_ldap/ldap_module.py
RouganStriker: Just following suit with the rest of the file. @emintham do you remember why we were inlining ldap imports originally?

emintham: I don't... Does it not work on top?

RouganStriker: I _think_ it should, will try after this test run.

RouganStriker: Yup still works, moved to top of file.

RouganStriker:
Will merge after my test run is complete

PR #516, opened 2017-06-26
pass in 'analysis_validate_cert' to request.

lmontopo:
The request which downloads the file from the server via http
should also have the config['analysis_validate_cert'] variable
passed in, to allow the option of turning off ssl cert validation.
Also add error handling for when no project is found in soap
response.

Refs: #SUP-69


sdetools/modules/import_fortify/fortify_ssc_importer.py
```diff
@@ -180,7 +182,7 @@ def run(self):
         req = urllib2.Request('%s/download/currentStateFprDownload.html?id=%s&mat=%s&clientVersion=3.60.0065' %
                               (self.ssc_endpoint, project_version_id, token))
 
-        opener = http_req.get_opener(self.config['analysis_method'], self.config['analysis_server'])
+        opener = http_req.get_opener(self.config['analysis_method'], self.config['analysis_server'], self.config['analysis_validate_cert'])
```
RouganStriker: I think the cert validation param might not be the 3rd param, we might want to specify it as a keyword argument `validate_certs=self.config['analysis_validate_cert']`

RouganStriker:
I did a quick search, it seems jira_soap.connect_server is another place where we aren't passing in the validate cert flag.

sdetools/modules/import_fortify/fortify_ssc_importer.py
lmontopo: ok cool thats a good idea just to be safe.

PR #520, opened 2017-07-11
Fixes in Mingle Plugin Meta-data

lmontopo:
Some small fixes to make this consistent with alm v1's copy /
scope of fields.


sdetools/modules/sync_mingle/mingle_metadata.py
```diff
@@ -25,28 +25,31 @@ def supports_references(self):
     def __init__(self, phases):
         super(MingleMetadata, self).__init__(phases)
 
-        # update server field
-        self.alm_server.default = ''
```
sibartlett: We should have the following placeholder: `your-company.mingle-api.thoughtworks.com`

lmontopo: done :)

sdetools/modules/sync_mingle/mingle_metadata.py
```diff
@@ -25,28 +25,32 @@ def supports_references(self):
...
+        # update placeholder of alm_server
```
sibartlett: Do we need these comments? I feel like the code already describes itself.

lmontopo: I think you're right actually. gonna remove em.

sibartlett:
Good stuff!

PR #521, opened 2017-07-11
Updates to RallyMetaData for consistency.

lmontopo:
Make copy similar to current copy.


sdetools/modules/sync_rally/rally_metadata.py
```diff
@@ -25,38 +25,48 @@ def supports_references(self):
     def __init__(self, phases):
         super(RallyMetadata, self).__init__(phases)
 
-        # update server field
-        self.alm_server.default = ''
```
lmontopo: Note to self: we gotta figure out what should be here.

PR #522, opened 2017-07-11
Remove redundant "json Field" labels.

lmontopo:
User no longer needs to input in JSON format.
Also change scope of certain ALM fields to correspond with current
behaviour.


sdetools/modules/sync_jira/jira_metadata.py
```diff
@@ -33,32 +36,39 @@ def __init__(self, phases):
             label=_('Jira version'),
             field_type=FieldType.RADIO,
             default='5',
-            options=[Option('4', '4'),
-                     Option('5+', '5')]
+            options=[
+                Option('4', '4'),
+                Option('5+', '5')
+            ],
+            scope=['global']
```
sibartlett: Scope are called `system`. Same below.

lmontopo: oups! i'm being sloppy.  Thanks for catching this.

PR #524, opened 2017-07-11
Verification of Github

lmontopo:


sdetools/modules/sync_github/github_metadata.py
```diff
@@ -25,33 +25,28 @@ def supports_references(self):
     def __init__(self, phases):
         super(GithubMetadata, self).__init__(phases)
 
-        # update alm project field
         self.alm_project.help_text = _('Repository name')
-        self.alm_project.description = self.alm_project.help_text
```
lmontopo: We only need one "Repository name" text (otherwise we see duplicate in the UI), but should it be `help_text` or `description`?  I kept it as `help_text`, but i wasn't sure.

sibartlett: I think you should change the label instead.

sdetools/modules/sync_github/github_metadata.py
```diff
@@ -25,33 +25,28 @@ def supports_references(self):
...
-
-        # update token field
         self.token.help_text = _('Go to "Personal Access Tokens" under your GitHub settings.')
 
         self.github_repo_owner = IntegrationField(
             name='github_repo_owner',
             label=_('Owner of this GitHub Repository'),
-            field_type=FieldType.STRING
+            field_type=FieldType.STRING,
+            scope=['project']
```
sibartlett: This should be system, at least that's what it currently is.

lmontopo: you are right! my bad! thanks for catching this!

sdetools/modules/sync_github/github_metadata.py
lmontopo: ah ok cool, good idea.

PR #528, opened 2017-07-12
Verification of TFS

Gee19:
* Remove useless password help text from base plugin
* Add area path & collection name fields
* Add advanced options section to system level
* Add missing scope to TFS fields
* Append tfs_collection field to connection details section
* Show area path & iteration path fields on project level under
Advanced TFS Configuration section.

Refs #SDE-3970


sdetools/modules/sync_tfs/tfs_metadata.py
```diff
@@ -25,43 +25,74 @@ def supports_references(self):
     def __init__(self, phases):
         super(TfsMetadata, self).__init__(phases)
 
+        self.alm_server.placeholder = 'bugs.server.com'
+
```
lmontopo: whats the decision here @simon? Are we going with "bugs.server.com" whenever there isn't an alm specific placeholder to put? Or are we leaving it blank in such cases?

sdetools/modules/sync_tfs/tfs_metadata.py
```diff
@@ -25,43 +25,74 @@ def supports_references(self):
...
             label=_('States which map to DONE in SD Elements'),
             field_type=FieldType.STRING,
-            placeholder='Done',
-            default='Done'
+            default='Done',
+            scope=['system']
         )
         self.tfs_new_status = IntegrationField(
             name='tfs_new_status',
             label=_('Open a story in %s with the following status' % self.name),
```
lmontopo: Our wording should be consistent.  Here we are using "open a _story_", but below (for the reopen status) we use "reopen a _work item_".  I don't have any preference between _work item_ and _story_ i just think we should pick one and use it for both fields.  Thoughts?

lmontopo:
Everything looks good, just a couple small questions / suggestions.

sdetools/modules/sync_tfs/tfs_metadata.py
Gee19: I definitely think we should use the proper terminology for the specific ALM. Fixing now

lmontopo: what is the property terminology? work item or story?

sdetools/modules/sync_tfs/tfs_metadata.py
Gee19: Removed it, it was set as a placeholder in the base class.

PR #531, opened 2017-07-12
Use .format() instead of %s for string formatting.

lmontopo:
Remove redundant comments on top of field changes.

Consistently kept one line in between inherited field changes
for easier readability.

Use `dedent` so that markdown dependant text can be broken up onto 
multiple lines.


sdetools/modules/sync_github/github_metadata.py
```diff
@@ -1,5 +1,8 @@
+from textwrap import dedent
+
 from sdetools.alm_integration.alm_metadata_plugin_base import (BaseALMMetadata,
-                                                               FieldType, IntegrationField)
+                                                               FieldType,
+                                                               IntegrationField)
```
Gee19: We should do this to all the plugins if this is the style we want

lmontopo: op, yea we should. maybe i'll just undo this change for now, cause i'm not actually sure.

lmontopo: actually, no, i do like it this way better, so i will change each one.

PR #532, opened 2017-07-12
Fix TravisCI Tests

Gee19:


sdetools/modules/sync_pt/pt_metadata.py
```diff
@@ -78,6 +78,3 @@ def get_advanced_alm_config_section(self):
             self.pt_group_label
         ])
         return section
-
-    def get_advanced_options_section(self):
-        return None
```
lmontopo: Ah ok this can't be removed though! cause by default this will return things, which will make pt have an advanced sync section even though its not supposed to!

lmontopo: I was afraid that this is what the failed test was about. Wasn't sure though.

Gee19: It's already defined above

lmontopo: It shouldn't be defined above AFAIK @jordan.  Cause PT doesn't have an advanced options section.  Unless we are adding one?

Gee19:
I messed up this MR, gonna close & re-open once leta's MR is merged.

PR #535, opened 2017-07-13
Fix advanced options section ordering

Gee19:
extendleft reversed the order of the fields in the advanced
options section, this reverts them all back to match the old UI.


Gee19:
Merging this because I want to verify some plugins, it's annoying trying to verify when all the fields are reversed.

PR #539, opened 2017-07-14
Added new ALM metatdata plugins

sibartlett:
We've added new ALM metadata plugins, that describe the following:

* Plugin ID
* Plugin name (and short name if applicable)
* Command
* Whether they support task references
* Lists sections and fields that are to be rendered in the create/edit forms in SDE

Currently, these plugins only provide data that's needed by SDE - but in the future, could potentially be extended to provide other metadata.

NOTE: Making ALM integrations fully-pluggable is outside the scope of this pull request. We're only concerned with making the integrations self-describing, to reduced the coupling between sdetools and SDE.

People who made this happen: @lmontopo, @Gee19 and myself.


sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,613 @@
...
+    """
+    Represents a Section of fields in the UI.
+    Example: username and password fields are in the Section with label
+    'Credentials'.
+    """
+
+    def __init__(self, label=None, optional=False, fields=None, description=None):
+        self.label = label
+        self.optional = optional
+        self.fields = list(fields or [])
```
lmontopo: do we still need this line to be this way? won't `self.fields=fields` do the trick now we'll be passing in a list?

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,613 @@
...
+        pass
+
+    @property
+    def short_name(self):
+        """ Returns a shortened name for the integration """
+        return self.name
+
+    @abstractproperty
+    def command(self):
+        """ Return a command name used for syncing the specific integration """
```
lmontopo: maybe we should be more consistent with our docstrings.  Sometimes we say "Return ..." sometimes we say "Returns ..." And should there be a period at the end of our docstrings? Im going to read PEP 257 then come back to this.

lmontopo: K so according to PEP 257 we should use Return instead of Returns and we should add periods.  I'll do that and make a pr.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,613 @@
...
+        Option('[$application - $project] $task_id $title', _('[Application - Project] T1: Task title')),
+        Option('[$application - $context] $task_id $title', _('[Application - Context] T1: Task title')),
+        Option('[$context] $task_id $title', _('[Context] T1: Task title'))
+    ]
+
+    PHASE_OPTIONS = [
+        Option('requirements', 'Requirements'),
+        Option('architecture-design', 'Architecture & Design'),
+        Option('development', 'Development'),
+        Option('testing', 'Testing')
```
Gee19: Should these be internationalized?

`Option('requirements', _('Requirements'))`

lmontopo: ohhh good point I think they should be.

sibartlett: I guess they could, but the SDE content doesn't currently support i18n - hence why I skipped these. They're support to mirror the phases which are part of the content.

sdetools/alm_integration/alm_metadata_plugin_base.py
sibartlett: I pass it into a list for good measure, if a future developer decides to pass a tuple or something else in - it'll get converted to a list.

lmontopo: oh ok true. Thanks for clarifying. :)

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+            help_text=_(
+                "A customized set of fields to add to new issues during a "
+                "synchronization.\n\n{0}"
+            ).format(self.DYNAMIC_MACROS),
+            features=['advanced_alm']
+        )
+        self.alm_issue_label = IntegrationField(
+            name='alm_issue_label',
+            label=_('Issue Labels / Tags'),
+            placeholder='SD-Elements',
```
shawngustaw: No description / help text?

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+            scope=['system'],
+            field_type=FieldType.STRING
+        )
+        self.alm_parent_issue = IntegrationField(
+            name='alm_parent_issue',
+            label=_('Alm parent issue'),
+            field_type=FieldType.STRING,
+            scope=['project'],
+            description=_(
```
shawngustaw: This seems like something each ALM should implement and throw an exception if they don't.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+        self.alm_custom_lookup_fields = IntegrationField(
+            name='alm_custom_lookup_fields',
+            label=_('Custom Lookup Fields'),
+            field_type=FieldType.JSON,
+            max_length=1024,
+            scope=['system'],
+            deprecated=self.supports_references,
+            description=_(
+                "Method to sync more than one SD Elements "
+                "project to a {0} project"
```
shawngustaw: Is this what custom lookup fields do? If you asked me I would never have said anything to do with syncing more than 1 project.

sdetools/modules/i18n.py
```diff
@@ -0,0 +1,5 @@
+try:
+    from django.utils.translation import ugettext_lazy as _
```
shawngustaw: Doesn't this add Django as a dependency? Afaik the goal of SDETools is to be free of dependencies.

sdetools/modules/sync_pt/pt_metadata.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class PtMetadata(BaseALMMetadata):
+    """
+    Provides metadata for Pivotal Tracker.
+
+    Most functionality is inherited from its base class while overriding
+    certain methods to add additional fields specific to Pivotal Tracker.
+    One section is overwritten to remove default options not supported in
+    Pivotal Tracker.
```
shawngustaw: I don't think that last line is necessary. Could probably leave it as a docstring on the method itself that returns None.

sdetools/modules/i18n.py
sibartlett: That’s why it’s in a try block, the except falls back to a non-django stub method.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+    BOOLEAN = 'boolean'
+    MASKED_STRING = 'masked_string'  # For sensitive fields
+    CHECKBOX = 'checkbox'
+    RADIO = 'radio'
+    SELECT = 'select'
+    JSON = 'json'
+
+
+class Option(object):
+    """ Represents an Option of an IntegrationField with select FieldType."""
```
lmontopo: Question: do you think something like `Option('DONE', 'done')` is enough of an example, or should the docstring provide an example of an entire field using many `Option`s? Like 
``` 
self.sde_statuses_in_scope = IntegrationField(
            name='sde_statuses_in_scope',
            label=_('Tasks with status'),
            field_type=FieldType.CHECKBOX,
            default='TODO',
            options=[
                Option('DONE', 'Done'),
                Option('TODO', 'Todo'),
                Option('NA', 'N/A')
            ],
            scope=['system']
)
```

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+
+class BaseALMMetadata(object):
+    """
+    Abstract base class to represent an ALM integration.
+
+    Subclassed by implementations for specific plugins that can translate the
+    required ALM fields to SDE fields.
+    """
+
+    STANDARD_STATUS_LIST = ['TODO', 'DONE', 'NA']
```
lmontopo: I do agree. I am hesitant to change this just because we have the same macro that we use in other SDE Tools files, and it is named the same way.  So on the one hand, yes, its a bad name, on the otherhand, should we be consistent?  If, at some point, the metadata file merges with the main plugin (sync) file, it'd be good to have names the same so that the merge is easier.  Or we could just go through and change the name always to be a better name... just not sure if its worth the effort or not.  But if you think it is, or if you think its OK to just have the macros be different in different files, i'll do it :)

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+            field_type=FieldType.BOOLEAN,
+            default=False,
+            scope=['system']
+        )
+        self.title_format = IntegrationField(
+            name='title_format',
+            label=_('ALM Title Format'),
+            field_type=FieldType.SELECT,
+            description=_(
+                "Customize the issue titles created in the "
```
lmontopo: good catch, thanks! fixed.

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: I'm not sure i understand what you mean.  Maybe you're saying that, like our `abstractproperty` on things like `name`, this should be turned into something that the developer making a new ALM Plugin should have to create?  Is that what you mean? If so, I am not sure. I suppose I don't see anything wrong with making them implement this field, but the current behaviour (where each plugin automatically inherits it) also seems to work.  And it requires less work for the devs. Perhaps your pointing out, though, that we've made it possible to override this, / not include it in the final list of fields / sections, and you're saying that we should have something to catch if a developer has done this, and if so, raise an error?

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: Yea, same. @sibartlett, @Gee19  can you confirm that this description is actually correct?

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+        self.description = description
+        self.max_length = max_length
+        self.required = required
+        self.placeholder = placeholder or default
+        self.options = options or []
+        self.features = features or []
+        self.scope = scope or []
+        self.deprecated = deprecated
+
+    def is_required(self, scope, parent_value=None):
```
lmontopo: good call. I've done this, when i push please take a look and let me know if it suffices.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+            field_type=FieldType.MASKED_STRING,
+            scope=['system']
+        )
+        self.token = IntegrationField(
+            name='alm_api_token',
+            label=_('{0} Token').format(self.short_name),
+            field_type=FieldType.STRING,
+            required=True,
+            scope=['system']
+        )
```
lmontopo: No, they all require either u/p OR token.  So by default each Plugin instance has all 3 fields, but the plugin has to create a credentials field, and we've created helper methods that create a section with the required fields.  

The way to not have all 3 be on each plugin, I suppose, would be to create them IN the helper methods `create_token_credentials_section`and `create_username_password_credentials_section`.  Maybe thats better, actually.  Is this what you were suggesting?

lmontopo: only thing actually about creating them then as opposed to in the `__init__` is that it makes it more complicated (I think) for the user to customize the token or username/password field.  If the token is called a `key` somewhere, and the token field isn't created in the `__init__` method, then they'd have to completely overwrite the helper method to change the copy, wouldn't they?

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+        Option('10', _('10 - High Minimum Priority to Import')),
+        Option('9', _('9')),
+        Option('8', _('8')),
+        Option('7', _('7')),
+        Option('6', _('6')),
+        Option('5', _('5 - Medium Minimum Priority to Import')),
+        Option('4', _('4')),
+        Option('3', _('3')),
+        Option('2', _('2')),
+        Option('1', _('1 - Low Minimum Priority to Import'))
```
lmontopo: Yea that may be an AA question. Will look into it tomorrow.

sdetools/alm_integration/alm_integration_plugin_service.py
```diff
@@ -0,0 +1,32 @@
...
+        PtMetadata,
+        RallyMetadata,
+        RationalMetadata,
+        TfsMetadata,
+        TracMetadata,
+        VersionOneMetadata
+    ]
+
+    @classmethod
+    def get_plugins(cls, phases=None):
```
lmontopo: good call, done.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+        self.alm_priority_map = IntegrationField(
+            name='alm_priority_map',
+            label=_('Custom Priority Mapping'),
+            field_type=FieldType.JSON,
+            max_length=1024,
+            scope=['system'],
+            description=_(
+                "If you need to set up a custom priority mapping please get in "
+                "touch with SD Elements support for help on filling "
+                "in the following field."
```
lmontopo: Totally Agree. I'll have to look into what Custom Priority Mapping actually is to be able to write a better description, unless @simon knows off the top of his head :)

sibartlett: It's a JSON object, where the keys are SDE priorities, and the values are ALM priorities.

Example:

```json
{
  "1-3": "Low",
  "4-6": "Medium",
  "7-9": "High",
  "10": "Critical"
}
```

sdetools/alm_integration/alm_metadata_plugin_base.py
sibartlett: Yep, this is a way to sync more than 1 SDE project to an ALM project. We could try coming up with a better description, but this field is mostly deprecated so I don't think it needs to be perfect.

sdetools/alm_integration/alm_metadata_plugin_base.py
sibartlett: Description and help text are optional.

sdetools/alm_integration/alm_metadata_plugin_base.py
sibartlett: Where is this used? I can not find any usage in this pull request?

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+        Option('4', _('4')),
+        Option('3', _('3')),
+        Option('2', _('2')),
+        Option('1', _('1 - Low Minimum Priority to Import'))
+    ]
+
+    DYNAMIC_MACROS = _("Field values can use the following dynamic macros. Their "
+                       "values change depending on the project being synced.\n\n"
+                       "* ${application} - Application name\n"
+                       "* ${project} - Project name\n"
```
sibartlett: @geoffwhittington thinks we should remove the macros from this help text

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: ha! well then we should get rid of it! its a macro in some of the sync files, but if its not here, i'll remove it. (i'll double check first of course)

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: @shawngustaw do you think that this field is confusing and needs description or help text?  I thought it was one of the more straight forward fields and didn't need any explanation, but if you disagree let us know.

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: Ok cool, since this description isn't outright wrong, i'm cool leaving it as is.  Thanks for confirming @simon.

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: yup, not used. removed.

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: so wait, @geoffwhittington things the entire `DYNAMIC MACROS` part should be omitted?

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: OK, updated the description. :)

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,623 @@
...
+        )
+        self.context_root = IntegrationField(
+            name='alm_context_root',
+            label=_('Context Root'),
+            field_type=FieldType.STRING,
+            placeholder='/',
+            scope=['system'],
+            help_text=_(
+                'Normally left empty unless the service is installed at a '
+                'non-standard location'
```
lmontopo: I see what you're saying @shawngustaw but i also think we are telling them when they might need this: when they have a service installed at a non standard location.  Perhaps we could include info telling them that this comes from the URL?  @sibartlett @geoffwhittington any thoughts?

sdetools/alm_integration/alm_metadata_plugin_base.py
sibartlett: Yeah, eventually we'll direct users to documentation for all the available macros - there's a lot more macros than the examples we have here.

sdetools/alm_integration/alm_metadata_plugin_base.py
sibartlett: If we add help text to say it comes from the URL, we might want to do it on a per-ALM basis - as the finer details are slightly different between ALMs.

lmontopo: So you mean that it doesn't always come from the URL? Or you mean that it does always come from the URL, but where in the URL depends on the ALM?

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: OK cool thanks for confirming.  I've removed the macros.

sdetools/alm_integration/alm_integration_plugin_service.py
shawngustaw: @lmontopo Looking at this today it's kind of obvious what the code does but I guess I'm just confused as to how the plugins and phases actually relate to each other.

sdetools/alm_integration/alm_metadata_plugin_base.py
shawngustaw: Makes a *ton* of sense now. Thanks @lmontopo !

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -0,0 +1,639 @@
...
+        :param parent_value : The value of this field on a parent connection,
+                              if applicable.
+        """
+        if not self.required:
+            return False
+
+        if scope == 'system' and 'project' in self.scope:
+            return False
+
+        if scope == 'project' and 'system' in self.scope and parent_value:
```
shawngustaw: This might be a good place for a human readable comment as well. It takes a couple seconds to reason about this line. Up to you though!

sdetools/alm_integration/alm_metadata_plugin_base.py
shawngustaw: @lmontopo  Yeah I guess I'm just wondering what happens to `self.username` and `self.password` when a Plugin only uses `self.token`.

sdetools/alm_integration/alm_metadata_plugin_base.py
shawngustaw: I don't see the harm in adding one. Being explicit is a very good habit imo

sdetools/alm_integration/alm_metadata_plugin_base.py
shawngustaw: @lmontopo I was suggesting it be an `abstractproperty` which would exception out if not implemented. I think at bare minimum each ALM should override `self.alm_parent_issue.description` with the ALM specific stuff and explaining how to use it... 

I feel like the whole point of doing this is so you don't have to say "Please get in touch with us".

sdetools/alm_integration/alm_metadata_plugin_base.py
shawngustaw: I remember vaguely now how this works. If something is marked as deprecated will it show up? If no one ever see's this then don't worry about it but if someone's on an older version it probably makes sense to give them a better description/example

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: They are just there, but aren't used. I think thats OK, do you think its OK @shawngustaw ? Is it confusing? If so maybe we could add a comment somewhere mentioning this?

sdetools/modules/sync_pt/pt_metadata.py
shawngustaw: I actually don't think the second paragraph in any of the plugin docstrings is actually that helpful. 

If you just say `Provides metadata specific to Pivotal Tracker.` then you have the exact same information as the entire 2-3 sentences.

sdetools/alm_integration/alm_metadata_plugin_base.py
Gee19: We are only adding description and help text if they add sufficient value for the end user, otherwise the UI becomes too cluttered.

The form needs to be helpful but also clear and concise.

PR #541, opened 2017-07-14
Update docstring formatting, add missing docstrings.

lmontopo:
Add docstrings to each ALM's Metadata class.
Make docstring formatting consistent.


sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -162,9 +162,10 @@ def run_validation(self, scope, value, parent_value=None):
 
 class BaseALMMetadata(object):
     """
-    Abstract base class to represent an integration. This should be subclassed
-    by an implementation for a specific plugin that can translate the required
-    ALM fields to SDE fields.
+    Abstract base class to represent an integration.
```
sibartlett: Missing "ALM", should be:

> Abstract base class to represent an ALM integration.

lmontopo: true.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -436,29 +437,29 @@ def __init__(self, phases):
...
-        """ Return a command name used for syncing the specific integration """
+        """ Return a command name used for syncing the specific integration. """
         pass
 
     @abstractproperty
     def supports_references(self):
         """
-        Return Boolean of whether alm supports references
-        (including task refrences)
+        Return Boolean of whether the alm supports references
```
sibartlett: alm should be uppercase

sdetools/modules/sync_github/github_metadata.py
```diff
@@ -5,6 +5,13 @@
 
 
 class GithubMetadata(BaseALMMetadata):
+    """
+    Provides metadata for Github.
+
+    Most functionality, including validation, is inherited from its base class.
+    Extends get_connection_details_section, get_advanced_options_section
+    and get_advanced_alm_config_section to add additional fields.
```
sibartlett: This docstring is incorrect, it doesn't match the other ALMs.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -436,29 +437,29 @@ def __init__(self, phases):
 
     @abstractproperty
     def id(self):
-        """ Returns an ID string compatible with SD Elements """
+        """ Return an ID string compatible with SD Elements. """
```
sibartlett: How about this?

> Return a string, that uniquely identifies the ALM integration.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -436,29 +437,29 @@ def __init__(self, phases):
...
         pass
 
     @abstractproperty
     def name(self):
-        """ Returns a name for the integration """
+        """ Return a name for the integration. """
```
sibartlett: > Return a string, that represents the product name of the ALM.

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: yes that is better.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -436,29 +437,29 @@ def __init__(self, phases):
...
         pass
 
     @property
     def short_name(self):
-        """ Returns a shortened name for the integration """
+        """ Return a shortened name for the integration. """
```
sibartlett: > Return a string, that represents a shortened or abbreviated version of the product name.

sdetools/alm_integration/alm_metadata_plugin_base.py
sibartlett: > Return Boolean of whether the integration supports SDE Task References.

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -436,29 +437,32 @@ def __init__(self, phases):
...
+        """ Return a command name used for syncing the specific integration. """
         pass
 
     @abstractproperty
     def supports_references(self):
         """
-        Return Boolean of whether alm supports references
-        (including task refrences)
+        Return Boolean of whether the ALM supports references
+        (including task refrences).
```
Gee19: spelling mistake

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -436,29 +437,32 @@ def __init__(self, phases):
...
-        """ Return a command name used for syncing the specific integration """
+        """ Return a command name used for syncing the specific integration. """
         pass
 
     @abstractproperty
     def supports_references(self):
         """
-        Return Boolean of whether alm supports references
-        (including task refrences)
+        Return Boolean of whether the ALM supports references
```
sibartlett: This should be:

> Return Boolean of whether the integration supports SDE Task References.

Because it the integration that supports references - NOT the ALM itself.

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: yup! thanks!

sdetools/alm_integration/alm_metadata_plugin_base.py
lmontopo: hm... not sure i understand this. What do you mean "its the integration not the ALM itself".  Are you saying that we just haven't updated sdetools for a couple ALM's?

sibartlett: I mean only SDE and sdetools knows what task references are. ALMs, such as JIRA, don't know what a task reference is.

So "whether the ALM supports references" is technically incorrect, because it's the integration (not the ALM) that supports task references.

lmontopo: OK right, but am i still correct in thinking that the reason PT and Trac integrations with SDE don't support task references is because its impossible to implement it with those ALMs? Or is that incorrect?

lmontopo: (i'll change the docstring, but now i just want to make sure i actually understand this stuff.)

sibartlett: HP ALM and Trac don't support task references, because we haven't implemented it yet.

**Task References**
Once we sync a task to JIRA, we update the SDE task to include a "reference" (which is the JIRA ticket number and URL). In the future we'll display these as links on the task page, so the user can click through to the ALM.

lmontopo: That clarifies it.  Thank you :)

sibartlett: Oh, to add to my last comment - we use task references to day, during subsequent syncs. Instead of looking up a task in JIRA using it's title, we look it up using the JIRA ticket ID.

lmontopo: 👍  :)

sdetools/alm_integration/alm_metadata_plugin_base.py
```diff
@@ -101,6 +103,9 @@ def __init__(self, name, label='', description='', options=None,
 
         :param features: Lists features that are required to set this field.
                          Defaults to empty list.
+
+        :param deprecated: Specifies whether this field is still supported.
```
sibartlett: The current comment implies that `True` means supported, `False` means deprecated. I think it'd read better as:

> Specifies whether this field has been deprecated.

lmontopo: oh yea, true. I was trying to avoid using the word deprecated to define deprecated, but i think that your comment is better.

PR #542, opened 2017-07-25
Updated ALM sync, so tasks out of scope are closed

sibartlett:


geoffwhittington:
Nice refactor, it looks a lot clearer!

sibartlett:
I've updated HP ALM, so it now overrides `alm_get_tasks` instead of `filter_tasks`.

sibartlett:
I still got more work to do here lol

PR #543, opened 2017-07-31
Feature/custom status mapping

keyvanm:
- ~Works with 3 SDE statuses: NA, TODO and DONE~ Now works with all statuses, including custom ones.
- Tested on JIRA, ALM as authoritative source
- JIRA, SDE as authoritative source needs transitions. Seems to be doable with simple statuses and the default workflow.
- Tested on Rally, 2 way sync works


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -838,34 +845,40 @@ def _reference_out_of_date(self, old_reference, new_reference):
         changes = old_reference != new_reference
         return changes
 
+    def _set_alm_to_sde_custom_status_map(self):
+        status_map = {}
+        if self.config.get('custom_status_mapping'):
+            self.config.process_json_dict('custom_status_mapping')
+            custom_status_mapping = self.config.get('custom_status_mapping')
+            for value in custom_status_mapping.values():
+                if not (value == 'DONE' or value == 'TODO' or value == 'NA'):
```
funkaoshi: You can make a union of the `DONE_SET`, `TODO_SET`, and `NA_SET` sets and then confirm `value` is in that set.

keyvanm: It seems like although there can be any statuses, they can only have one of the 3 `meanings`: `DONE`, `TODO` and `NA`. That's great! It means that any status is going to be in one of `STATUS.NA_SET`, `STATUS.DONE_SET`, or `STATUS.TODO_SET`. I think this approach might work!

keyvanm:
I'm having a issue with setting the status of a task to one of the custom statuses, `CTS1`: In Progress. I have no issue with other custom statuses. 

```
{
	"id": "CTS1",
	"meaning": "TODO",
	"icon": "code",
	"name": "In Progress",
	"slug": "INPROGRESS",
	"requires_comment": false
}
```

This is the error I get: Unable to update the task status in SD Elements. Either the task no longer exists, there was a problem connecting to the server, or the status was invalid.

Any idea why this is happening?

geoffwhittington:
@sibartlett  @funkaoshi can we please close this pull request now if we're done with it?

PR #565, opened 2017-08-16
Add ability to sync Training content

shawngustaw:
Refs #SDE-4505


shawngustaw:
Some assumptions I've made: 

- Links should be in markdown format 
- Including Training content won't impact performance too much (I'm going to investigate making include filters dynamic but I don't want to get stuck refactoring a ton of code).

Gee19:
+1

geoffwhittington:
Training is valid only if there is a license. Will this change will add the Boolean checkbox for every instance regardless of whether or not training is available? It might be misleading to people.

If so, we might want to feed in the context like we did with phases?

shawngustaw:
@geoffwhittington If an org doesn't have a license the endpoint just returns an empty list so there shouldn't be an issue with unlicensed content bleeding through (:+1:).

I'm not sure how phases are being passed into SDETools though - if it hits an endpoint the caveat is we don't have an endpoint for licenses. I was thinking for now we could update the Field to say something like `(requires valid training license)` and then investigate whether we want to improve the UX.

sibartlett:
+1

sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -717,6 +720,15 @@ def sde_get_task_content(self, task):
                 content += '## %s\n\n' % (implementation['title'])
                 content += implementation['text'] + '\n\n'
 
+        if self.config['training_in_scope'] and task.get('training'):
+            content += '\n\n## Training Modules\n\n'
+            for course in task['training']:
+                content += '### %s\n' % (course['title'])
+                for module in course['modules']:
+                    content += '  * [{0}]({1})\n'.format(module['title'],
```
RouganStriker: Vey found an issue where links weren't rendering in some ALMs; the syntax varies between ALMs so we can't assume the markdown syntax will work everywhere. 

I'm thinking we should create a helper method in all ALMs for rendering links in content. Alternatively we can just insert the URL here, I think most ALMs are smart enough to auto-render it as a link.

sibartlett:
Training module links don't display correctly in JIRA, because they're wrapped in a `{code}` block.

sdetools/alm_integration/alm_plugin_base.py
sibartlett: The indentation causes the list of links to be treated as a code block.

PR #568, opened 2017-08-22
Add support for syncing stories under a story

geoffwhittington:


sibartlett:
Do we need to fix the logic on [line 282](https://github.com/sdelements/sdetools/blob/f19e8ac1f5f38a031e847f401da6933e8cb0a346/sdetools/modules/sync_rally/rally_plugin.py#L282)? It assumes the parent is a HierarchicalRequirement -is that right?

geoffwhittington:
Yup, thats right - tasks or stories can live under a HierarchicalRequirement 

In a next iteration it would be nice to introspect the fields and allow people to set relationships by their ID. In this case fields WorkProduct and Parent are of type OBJECT with schema HierarchicalRequirement

PR #569, opened 2017-08-24
Filter out LDAP users with no email.

RouganStriker:
* Fix rendering of LDAP sync user creation errors

Refs #SDE-4538


shawngustaw:
+1

PR #572, opened 2017-09-18
Update jira_markdown.py

geoffwhittington:
We are presently converting \_\_bold text\_\_ to \__bold text_\_ incorrectly. It should be __bold text__


geoffwhittington:
I haven't tested this change, but it looks like it should work. Buyer beware!

Gee19:
+1

PR #574, opened 2017-09-25
Sync empty custom attribute field mapping values

Gee19:
If a custom attribute is optional and an ALM connection includes it in
the custom field mapping, sync an empty string instead of throwing an
exception.

Refs #SDE-4675


geoffwhittington:
:+1:

PR #577, opened 2017-09-28
Try to parse a copy of the xml file to avoid close

MartinPetkov:
The sax parser closes the file passed to it, so we pass a deep copy and then close it. Closing multiple times is fine, and we can't use `with` because StringIO doesn't have an __exit__ method.


sdetools/analysis_integration/base_integrator.py
```diff
@@ -292,7 +293,11 @@ def get_parse_was_successful(self):
 
     def can_parse_file(self, xml_file):
         try:
-            self.parse(xml_file)
+            # The sax parser will actually close the file, which we don't want
+            xml_file_copy = copy.deepcopy(xml_file)
```
emintham: Maybe you want `xml_file.dup()` here instead if you're duplicating file descriptors?

sdetools/analysis_integration/base_integrator.py
```diff
@@ -292,7 +293,11 @@ def get_parse_was_successful(self):
...
+            self.parse(xml_file_copy)
+            if hasattr(xml_file_copy, "close"):
```
emintham: I don't understand this. If `self.parse(xml_file_copy)` does the closing already, why are we calling the `close` method again?

sdetools/analysis_integration/base_integrator.py
MartinPetkov: Thank you, I'll look into using that.

MartinPetkov: xml_file in this case can be a StringIO, so I can't use `os.dup()` on it

sdetools/analysis_integration/base_integrator.py
```diff
@@ -292,7 +294,9 @@ def get_parse_was_successful(self):
 
     def can_parse_file(self, xml_file):
         try:
-            self.parse(xml_file)
+            # The sax parser will actually close the file, which we don't want
+            self.parse(copy.deepcopy(xml_file) if isinstance(xml_file, StringIO) else xml_file)
```
emintham: After reading the other parts of the code, I think I understand it a bit better..

The problem seems to be when we're passing xml_file as a StringIO instance. In that case, StringIO has `.seek` and `.close` and StringIO calls `.close` at the end of parsing it (which is actually recommended practice).

One idea is to add an `elif` clause to the `parse` method and pass it by value.

```python
def parse(self, file_name):
    # given name of file
    if isinstance(file_name, basestring):
        try:
            # no need to seek since this will be a new file pointer
            self.parse_file(open(file_name, 'r'))
        except IOError:
            raise WhateverError
    # given StringIO (string that acts like a file)
    elif isinstance(file_name, StringIO):
        self.parse_string(file_name.getvalue()) # this reads the entire contents of StringIO without needing to seek(0), it doesn't affect the "pointer" of the StringIO object
    # given an actual file pointer
    else:
        file_name.seek(0)
        self.parse_file(file_name)
```

MartinPetkov: Great suggestion! I'll give it a try tomorrow and report results.

sdetools/analysis_integration/base_integrator.py
```diff
@@ -267,10 +276,10 @@ def parse_file(self, xml_file):
 
         self.store_findings(XMLReader)
 
-    def parse_string(self, xml):
+    def parse_string(self, xml_string):
         XMLReader = self._get_content_handler()
         try:
-            sax.parseString(xml, XMLReader)
+            sax.parseString(xml_string, XMLReader)
```
MartinPetkov: I changed the name here because it was shadowing the "xml" package and the exceptions would break saying "<type> has no attribute 'sax'"

sdetools/analysis_integration/base_integrator.py
```diff
@@ -239,16 +240,24 @@ def _get_content_handler(self):
...
-        if isinstance(file_name, basestring):
+    def parse(self, file_obj_or_name):
+        # file_obj_or_name is the name of a file
+        if isinstance(file_obj_or_name, basestring):
             try:
-                fp = open(file_name, 'r')
+                self.parse_file(open(file_obj_or_name))
             except IOError as ioe:
-                raise IntegrationError("Could not open file '%s': %s" % (file_name, ioe))
+                raise IntegrationError("Could not open file '{0}': {1}".format(file_obj_or_name, ioe))
```
emintham: I think we want to keep the % string formatting here to support (really) *old* python versions. It's a pet peeve of mine as well but we don't get to change it :(

emintham:
Other than that, +1 if it works.

sdetools/analysis_integration/base_integrator.py
MartinPetkov: Oh okay then, I'll change it back

emintham:
+1

PR #578, opened 2017-10-03
Return empty string for custom attribute if it doesn't exist in mapping

Gee19:
Refs #SDE-4675


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -280,7 +280,7 @@ def _transform_value(self, value, mapping):
             macros = re.findall('\$\{?([a-zA-Z0-9_]+)\}?', value)
             for required_macro in macros:
                 if required_macro not in mapping:
```
geoffwhittington: I would call it `macro` since it is no longer required.

geoffwhittington: +1 otherwise

PR #579, opened 2017-10-03
Set mapping[macro] to empty string if value=null

Gee19:
Bump sdetools version to 5.12.9

Refs #SDE-4675


Gee19:
cc @hanifvirani

funkaoshi:
+1

PR #580, opened 2017-10-12
Validate ALM config parameters for connnections

Gee19:
Refs #SDE-4699


sdetools/alm_integration/alm_plugin_base.py
```diff
@@ -306,11 +306,11 @@ def alm_connect(self):
 
         self.alm_connect_project()
 
+        self.alm_validate_configurations()
```
geoffwhittington: I think we want this after `alm_connect_server()` since the values that this function is typically testing is set at the connector level.

Gee19: It doesn't seem to work for certain connectors at this level, for example rally is erroring out at line 316 of `rally_plugin.py` with a `list index out of range`:

`requirement_definition_ref = requirement_definition['QueryResult']['Results'][0]['_ref']`

